PR #14, opened 2022-02-10
Fix beaconstate participation fields in altair

alrevuelta:
After Altair, `previous_epoch_attestations` was renamed to `previous_epoch_participation` and `current_epoch_attestations` to `current_epoch_participation`.

This was causing that after fetching the `BeaconState`, the `beaconState.Altair.CurrentEpochAttestations` and `beaconState.Altair.PreviousEpochAttestations` fields were empty.
```console
beaconState, err := httpClient.BeaconState(context.Background(), "finalized")
if err != nil {
	panic(err)
}
log.Info(beaconState.Altair.CurrentEpochAttestations) --> Empty!
```

This PR fixes that issue by i) renaming the expected field from the fetched result from the node, and ii) renaming the fields to reflect Altair changes. 

Source: [Altair Spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#beaconstate)

### DIFF
--- a/spec/altair/beaconstate.go
+++ b/spec/altair/beaconstate.go
@@ -42,8 +42,8 @@ type BeaconState struct {
 	Balances                    []uint64             `ssz-max:"1099511627776"`
 	RANDAOMixes                 [][]byte             `ssz-size:"65536,32"`
 	Slashings                   []uint64             `ssz-size:"8192"`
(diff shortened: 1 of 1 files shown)


mcdee:
Thank you for your contribution!

PR #21, opened 2022-07-22
Add fallback if no active multi clients.

corverroos:
Adds the `WithFallback` parameter to `multi.Service` and fixes issue with `multi.doColl` that returns `nil,nil` when no active providers are available.

Fallback results in `doCall` to also attempt each inactive provider when no active providers are available. This seamlessly actives any actually active providers. This is quite common since providers are deactivated on a single error, which can be due to sporadic timeouts or other sporadic network issues.


corverroos:
Closing in favour of #22

PR #22, opened 2022-07-22
Recheck inactive multi clients as required.

### LINKED ISSUE #20: Issue when no active multi providers
When there are no active providers in a multi client, then `doCall` returns `nil,nil` which causes unexpected behaviour.

We at Obol encountered this after adding support for multiple beacon nodes using the `multi` package. Most of our tests and setups still only has a single beacon node. We configure our eth2 clients with a 2s timeout. This results in sporadic timeouts from beacon nodes in the wild (most calls are super fast, but sometimes around epoch change we suspect, calls sporadically take longer and timeout). This results in the only client being disabled.

Since these timeouts are sporadic, and subsequent calls will succeed, but we have only a single client, we are suggesting to allow "falling back" to inactive providers if no active providers are available. This will result in slower failures when all providers are actually down, but it will recover seamlessly if one of the "inactive providers" is actually "active", like in our case.  This is similar to AWS load balancers that "fail open" when all targets are unhealthy.

mcdee:
If a call is made and there are no active clients the inactive clients are immediately rechecked to see if they are now active.  If there are no active clients then an error is returned.

Fixes #20

### DIFF
--- a/multi/client.go
+++ b/multi/client.go
@@ -32,21 +32,26 @@ func (s *Service) monitor(ctx context.Context) {
 			log.Trace().Msg("Context done; monitor stopping")
 			return
 		case <-time.After(30 * time.Second):
-			// Fetch all clients.
-			clients := make([]consensusclient.Service, 0, len(s.activeClients)+len(s.inactiveClients))
-			s.clientsMu.RLock()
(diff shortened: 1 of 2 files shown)


corverroos:
LGTM 🚀

Issue #18, opened 2022-07-17
Builder API: Validator Registration


ciaranmcveigh5:
_To be used by Obols Charon client_

Validator clients have to register with a relay to gain access to the builder network. They are recommended to re-register every epoch.

The "validator registration" section reference in "Builder -- Honest Validator" has additional details

https://github.com/ethereum/builder-specs/blob/main/specs/validator.md#validator-registration

The specs for the payload that is sent with this registration can be found in the link below (ValidatorRegistrationV1 & SignedValidatorRegistrationV1)

https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#validatorregistrationv1

Happy to draft up a rough PR using "BlindedBeaconBlock" references in the repo as a template

Have made a start at implementing the changes on https://github.com/attestantio/go-eth2-client/pull/19

Thanks

PR #23, opened 2022-07-27
add Root function on versioned validator registration

ciaranmcveigh5:
- Adding Root functions for validator registration
- Adding DOMAIN_APPLICATION_BUILDER

Resolves Issues
https://github.com/attestantio/go-eth2-client/issues/18

### DIFF
--- a/http/spec_conformance_test.go
+++ b/http/spec_conformance_test.go
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build conformance
 // +build conformance
 
 package http_test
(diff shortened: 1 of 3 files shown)


http/spec_conformance_test.go
```diff
@@ -45,6 +46,7 @@ func TestSpecConformance(t *testing.T) {
 		"DOMAIN_RANDAO":                         phase0.DomainType{},
 		"DOMAIN_SELECTION_PROOF":                phase0.DomainType{},
 		"DOMAIN_VOLUNTARY_EXIT":                 phase0.DomainType{},
+		"DOMAIN_APPLICATION_BUILDER":            phase0.DomainType{},
```
ciaranmcveigh5, line 49: Not sure if you'd like a new folder called spec/v1 with file types.go in it for this and change phase0.DomainType{} to V1.DomainType{}

domain type reference can be found below

https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#validatorregistrationv1

ciaranmcveigh5: I also noticed this 

"Note: DOMAIN_APPLICATION_MASK reserves the rest of the bitspace in DomainType for application usage. This means for some DomainType DOMAIN_SOME_APPLICATION, DOMAIN_SOME_APPLICATION & DOMAIN_APPLICATION_MASK MUST be non-zero. This expression for any other DomainType in the consensus specs MUST be zero."

on https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md

which sounds like both DOMAIN_APPLICATION_BUILDER & DOMAIN_APPLICATION_MASK need to be set 

not sure if this has any impact on required code changes

PR #32, opened 2022-09-26
Remove race conditions

### LINKED ISSUE #31: Zerolog logger isn't threadsafe
We cannot run our tests with `-race` since it detects a race inside go-eth2-client. This is due to the use of a global `zerolog.Logger` that is not threadsafe. This global logger is used concurrently in the `http` package.

To reproduce, run: `go test -race ./...` in this repo.

mcdee:
Remove race conditions, notably from the logging system.

Fixes #31

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,3 +13,5 @@ jobs:
           go-version: 1.17
       - uses: actions/checkout@v3
       - uses: n8maninger/action-golang-test@v1
+        with:
+          args: "-race;-timeout=30m"
--- a/http/aggregateattestation_test.go
+++ b/http/aggregateattestation_test.go
@@ -26,6 +26,9 @@ import (
 )
 
 func TestAggregateAttestation(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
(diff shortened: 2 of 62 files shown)


corverroos (approved):

corverroos:
This test also passes, so looking good!
```

func TestConcurrentAccess(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	s, err := v1.New(ctx, v1.WithAddress(os.Getenv("HTTP_ADDRESS")), v1.WithTimeout(5*time.Second))
	require.NoError(t, err)

	const threads = 3

	var wg sync.WaitGroup
	for i := 0; i < threads; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()

			_, err := s.(client.SpecProvider).Spec(ctx)
			require.NoError(t, err)

			_, err = s.(client.NodeSyncingProvider).NodeSyncing(ctx)
			require.NoError(t, err)

			err = s.(client.EventsProvider).Events(ctx, []string{"head"}, func(*apiv1.Event) {})
			require.NoError(t, err)
		}()
	}

	wg.Wait()
}
```

PR #34, opened 2022-10-16
add structured http errors

corverroos:
Adds structured errors so users that do structured logging can extract fields and ensure consistent group-able error messages allowing proper metrics and alerts on errors.

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -36,6 +36,18 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+// Error represents an http error.
+type Error struct {
+	Method     string
+	Endpoint   string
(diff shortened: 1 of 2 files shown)


http/http.go
```diff
@@ -36,6 +36,18 @@ func init() {
...
+// Error represents an http error.
+type Error struct {
+	Method     string
+	Endpoint   string
+	StatusCode int
+	Data       []byte
+}
+
+func (e Error) Error() string {
+	return fmt.Sprintf("%s failed with status %d: %s", e.Method, e.StatusCode, e.Data)
```
corverroos, line 48: resulting error message exactly the same

PR #43, opened 2023-01-31
http: lowercase node version before comparing

xenowits:
Use `strings.Lower` to check client node versions. This solves the problem when clients send mixed case versions, for example, lighthouse returns `Lighthouse/v3.4.0-38514c0/x86_64-linux`. Also fixes typo in `nimbus`.

category: bug

### DIFF
--- a/http/nodeclient.go
+++ b/http/nodeclient.go
@@ -28,13 +28,13 @@ func (s *Service) NodeClient(ctx context.Context) (string, error) {
 	nodeVersion = strings.ToLower(nodeVersion)
 
 	switch {
-	case strings.HasPrefix(nodeVersion, "lighthouse"):
+	case strings.HasPrefix(strings.ToLower(nodeVersion), "lighthouse"):
 		return "lighthouse", nil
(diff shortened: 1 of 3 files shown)


mcdee:
`nodeVersion` is changed to lower case in line 28 of `nodeclient.go` so this does not need to happen again.

xenowits:
> `nodeVersion` is changed to lower case in line 28 of `nodeclient.go` so this does not need to happen again.

That's right, let me fix it

mcdee:
Those remaining `ToLower()`s still aren't necessary, because `nodeVersion` comes from `s.NodeVersion()` which carries out the `ToLower()` itself.

xenowits:
> Those remaining `ToLower()`s still aren't necessary, because `nodeVersion` comes from `s.NodeVersion()` which carries out the `ToLower()` itself.

Sorry but I can't find the `ToLower()` conversion in `s.NodeVersion()` implementation. If it's already there, then I can simply close the PR

mcdee (approved):

mcdee:
Sorry, was reading `NodeClient` not `NodeVersion`.

PR #45, opened 2023-02-16
Add support for extra HTTP headers in requests

avalkov:
For the purposes of services like Blockdaemon Ubiquity which require authentication, its general good idea to support extra headers in the requests.

### DIFF
--- a/http/http.go
+++ b/http/http.go
@@ -65,7 +65,9 @@ func (s *Service) get(ctx context.Context, endpoint string) (io.Reader, error) {
 		cancel()
 		return nil, errors.Wrap(err, "failed to create GET request")
 	}
+	s.addExtraHeaders(req)
 	req.Header.Set("Accept", "application/json")
+
 	resp, err := s.client.Do(req)
 	if err != nil {
 		cancel()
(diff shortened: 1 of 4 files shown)


http/http.go
```diff
@@ -66,6 +66,8 @@ func (s *Service) get(ctx context.Context, endpoint string) (io.Reader, error) {
 		return nil, errors.Wrap(err, "failed to create GET request")
 	}
 	req.Header.Set("Accept", "application/json")
+	s.addExtraHeaders(req)
```
mcdee, line 69: Please do this prior to the `req.Header.Set()` operations that are required, to avoid bad extra headers overwriting those set by the module (here and wherever else you do this).

http/http_test.go
```diff
@@ -34,3 +34,31 @@ func TestError(t *testing.T) {
 	require.Equal(t, nethttp.MethodGet, httpError.Method)
 	require.Equal(t, "/eth/v1/beacon/genesis", httpError.Endpoint)
 }
+
+func TestClientShouldSendExtraHeadersWhenProvided(t *testing.T) {
+	authroizationHeader := "Authorization"
```
mcdee, line 39: Please fix spelling here and elsewhere 'authroization' is present.

mcdee (changes requested):
Thank you for this PR.  Please could you address the highlighted items?  Thank you.

http/http.go
avalkov: Fixed.

http/http_test.go
avalkov: Fixed.

mcdee (approved):

PR #47, opened 2023-02-20
Add ssz hashing for transactions and withdrawals

avalonche:
Add utility functions to compute withdrawals and transactions roots ssz encoded.


mcdee:
I think that the best solution here is to rename `Transactions` to `ExecutionPayloadTransactions` and `Withdrawals` to `ExecutionPayloadWithdrawals`.  Not the most concise names, but given the specifics around the array sizes encoded in to the struct I don't want to give the impression that these are generic lists.

I'm also not a massive fan of these being in the `spec` directory, given that everything else under there is a consensus spec object.  I think these should be moved to `util/bellatrix` and `util/capella` directories, respectively.  The downside of this is that it will create multiple imports with the same name, but I think that's something we'll just have to live with (and already do so across repos anyway).

mcdee (approved):

PR #49, opened 2023-03-20
Add Debug Forkchoice endpoint support

samcm:
Adds support for the new ForkChoice debug endpoint.

I wasn't entirely sure on the structure/formatting of where to place these structs. Happy to be advised if there is a better way to do things 🙏 

Thank you!


http/forkchoice_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"os"
+	"testing"
+
+	client "github.com/attestantio/go-eth2-client"
+	"github.com/attestantio/go-eth2-client/http"
+	"github.com/stretchr/testify/require"
+)
+
+func TestForkChoice(t *testing.T) {
```
samcm, line 26: Little unsure on if this test is a good idea considering not all clients have implemented this endpoint yet and may disrupt any CI/CD/automated testing

mcdee:
This endpoint doesn't seem to be very well supported:

- lighthouse: no support
- nimbus: returns data format not as per the spec
- prysm: no support
- teku: support

so 1/4 of the nodes on which I tried it.  I'm a little concerned about adding support for this in to the codebase and then being pinged on issues using it due to the lack of downstream support.  Are there plans for the various clients to add/fix support?

samcm:
Yeah that's one of my concerns as well. Maybe it makes sense to rename the func to `DebugForkChoice` or something. As for support, yeah it's in flight with multiple clients:

Teku: Implemented
Lighthouse: Open PR https://github.com/sigp/lighthouse/pull/4003
Prysm: In `develop` - https://github.com/prysmaticlabs/prysm/commit/3d0ecdff3face5dceea47ac2c181d4d999903c3d
Nimbus: Behind `--debug-fork-choice` (I haven't validated the data though, you could definitely be correct that the format isn't per the spec)
Lodestar: Unfinalized implementation running on `/eth/v0/debug/forkchoice`

samcm:
Happy for this to stay unmerged until more support lands in client releases @mcdee FWIW 🙏

mcdee:
Yeah let's hold this until we see better coverage of support and can take a view on where it fits.

samcm:
@mcdee All clients excluding Lodestar have support in at least their `develop` branches now 🙏

api/v1/forkchoice.go
```diff
@@ -0,0 +1,234 @@
...
+	case "INVALID":
+		return ForkChoiceNodeValidityInvalid, nil
+	case "VALID":
+		return ForkChoiceNodeValidityValid, nil
+	case "OPTIMISTIC":
+		return ForkChoiceNodeValidityOptimistic, nil
+	default:
+		return "", errors.New(fmt.Sprintf("invalid value for fork choice node validity: %s", input))
+	}
+}
```
mcdee, line 108: This should ideally bit its own enum, similar to that in `spec/dataversion.go` with the individual types being uint64 and having the same functions.

mcdee (changes requested):
One change required.  Also, please could you add this call to the `sleepy` and `erroring` clients in the `testclients` directory?  Thanks.

api/v1/forkchoice.go
samcm: Have implemented that now @mcdee 🙏

mcdee:
There are a few errors being thrown up by the testing, please could you rebase this against he latest master and we'll see what needs to be addressed?  Thanks.

samcm:
Cheers @mcdee! Sorry I've been on holidays for a little bit, thank you for sorting that 🙏

PR #58, opened 2023-06-18
add ProposerIndex() to BlindedBlocks

GalRogozinski:
### Description

Added missing method


GalRogozinski:
@mcdee will you review?
Thanks in adavnce :-)

GalRogozinski:
solved conflicts

btw @mcdee, I noticed you added a `RandaoReveal()` to api/versionedblindedbeaconblock.go but not to spec/versionedbeaconblock.go. Is this on purpose?

mcdee:
Apologies for the delay on this.  I've been considering it, and my concern is that the interface is a subset of a beacon block rather than actually a beacon block.  So, for example, there is no `ExecutionPayload()` method which is something that a user seeing a `BeaconBlock` would reasonably expect to be able to obtain.

In general it's better to have interfaces with fewer functions, ideally 1.  Would it make sense to have interfaces as:

```
type BeaconBlockSlotProvider interface {
  Slot() (phase0.Slot, error)
}

type BeaconBlockProposerIndexProvider interface {
  ProposerIndex() (phase0.ValidatorIndex,error)
}

...
```

as that would allow a full range of interfaces that could cover both blinded and unblinded beacon blocks.

GalRogozinski:
I changed the description and title of the PR accordingly

PR #63, opened 2023-07-25
Add BlockHash() to versioned signed beacon block

avalonche:


spec/versionedsignedbeaconblock.go
```diff
@@ -66,6 +66,29 @@ func (v *VersionedSignedBeaconBlock) Slot() (phase0.Slot, error) {
 	}
 }
 
+// BlockHash returns the block hash of the beacon block.
+func (v *VersionedSignedBeaconBlock) BlockHash() (phase0.Hash32, error) {
```
mcdee, line 70: This should be `ExecutionBlockHash` as per the versioned signed blinded beacon block commit previously.

mcdee (changes requested):

mcdee (approved):

PR #67, opened 2023-07-31
Add versioned block requests

avalonche:
The [publishBlindedBlockV2](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/blocks/blinded_blocks.v2.yaml) and [publishBlockV2](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/blocks/blocks.v2.yaml) uses `BlockContents` for Deneb as request bodies.

This adds a versioned struct to support the new endpoints. Open to moving the structs to a more suitable name / location e.g. under a /v2 folder.

### DIFF
--- a/api/v1/deneb/blindedblobsidecar.go
+++ b/api/v1/deneb/blindedblobsidecar.go
@@ -26,7 +26,7 @@ type BlindedBlobSidecar struct {
 	BlockRoot       phase0.Root `ssz-size:"32"`
 	Index           deneb.BlobIndex
 	Slot            phase0.Slot
-	BlockParentRoot phase0.Root ` ssz-size:"32"`
+	BlockParentRoot phase0.Root `ssz-size:"32"`
 	ProposerIndex   phase0.ValidatorIndex
 	BlobRoot        phase0.Root         `ssz-size:"32"`
 	KzgCommitment   deneb.KzgCommitment `ssz-size:"48"`
--- a/api/versionedblindedblockrequest.go
+++ b/api/versionedblindedblockrequest.go
@@ -1,0 +1,308 @@
+// Copyright © 2023 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


avalonche:
my concern here is the json marshalling / unmarshalling that is not uniform across the endpoints (e.g. some endpoints have additional metadata such as `version` while some endpoints only contain the message). This makes it hard to unify the json marshalling without having specific logic for each endpoint and custom structs.
I think this needs to be resolved at the spec level to standardise a format containing {"metadata":..."data":...} for endpoints going forward. The versioned structs will need to have specific marshalling logic for each endpoint for the time being.

mcdee:
All endpoints in the beacon API have a `data` entity for the returned object(s), with optional and response-dependent metadata.  At current the codebase does not pay attention to any of the metadata items.  It would be nice to have all metadata returned as well, but that would be a breaking change.

The versioning would need to continue, but only at the stage where the JSON is being unmarshalled.  My comment was more focused on providing an easier interface for users and moving some of the difficulty into the codebase.  Something like:

```go
proposal := &deneb.SignedBlockContents{}
err := client.SubmitSignedProposal(ctx, proposal)
```

rather than the current:

```go
proposal := &spec.VersionedSignedBeaconBlock{
  Version: spec.DataVersionDeneb,
  Deneb: &deneb.SignedBlockContents{},
}
err := client.SubmitBeaconBlock(ctx, proposal)
```

And then `SubmitSignedProposal` could run a type switch to work out what it had been passed and handle accordingly.

Basically: I'm fine with the API returning `Versioned*` but thinking that we shouldn't be requiring users to create them to provide to the API.

avalonche:
What would the `consensusclient.SubmitResults` type look like? Does `submitBeaconBlock` also accept something more generic or just `spec.VersionedSignedBeaconBlock`?
From the standpoint of the submit block api this makes sense. However this is slightly different from my use case which is deserialising the struct from a json payload.

mcdee:
I'm going to merge this as-is, with the caveat that many of the function signatures are going to change in the next major release of this module so structs and signatures may be moved around at that point.

Issue #70, opened 2023-08-14
http service is leaking tickers


draganm:
ticker created here https://github.com/attestantio/go-eth2-client/blob/912094d609956fb3fcec679551b96e5ea22dbefc/http/service.go#L167 is never stopped, which leaves a dangling go-routine.

PR #71, opened 2023-08-14
stop refreshTicker on context cancel in http/service.go

draganm:
fix for #70: stop the ticker when the context is cancelled.


mcdee:
Thank you for this.

Fixes #70

PR #74, opened 2023-09-23
Add more unified getters for common block properties

pk910:
Added more unified getters for common block properties to the `VersionedSignedBeaconBlock` struct:
* ProposerIndex
* ExecutionBlockNumber
* ExecutionTransactions
* Graffiti
* RandaoReveal
* ETH1Data
* Deposits
* VoluntaryExits
* BLSToExecutionChanges
* Withdrawals
* BlobKzgCommitments

Fixed `VersionedSignedBeaconBlock.ExecutionBlockHash()`, which was using the bellatrix struct regardless of the actual block type and therefore fails with a nil pointer for non-bellatrix blocks

### DIFF
--- a/spec/versionedsignedbeaconblock.go
+++ b/spec/versionedsignedbeaconblock.go
@@ -66,6 +66,39 @@ func (v *VersionedSignedBeaconBlock) Slot() (phase0.Slot, error) {
 	}
 }
 
+// ProposerIndex returns the proposer index of the beacon block.
+func (v *VersionedSignedBeaconBlock) ProposerIndex() (phase0.ValidatorIndex, error) {
+	switch v.Version {
+	case DataVersionPhase0:
+		if v.Phase0 == nil {
+			return 0, errors.New("no phase0 block")
+		}
+		return v.Phase0.Message.ProposerIndex, nil
+	case DataVersionAltair:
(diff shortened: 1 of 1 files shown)


spec/versionedsignedbeaconblock.go
```diff
@@ -78,17 +111,98 @@ func (v *VersionedSignedBeaconBlock) ExecutionBlockHash() (phase0.Hash32, error)
...
+		if v.Deneb == nil || v.Deneb.Message == nil || v.Deneb.Message.Body == nil || v.Deneb.Message.Body.ExecutionPayload == nil {
+			return nil, errors.New("no deneb block")
+		}
+		return v.Deneb.Message.Body.ExecutionPayload.Transactions, nil
+	default:
+		return nil, errors.New("unknown version")
+	}
+}
+
+func (v *VersionedSignedBeaconBlock) Graffiti() ([]byte, error) {
```
mcdee, line 174: The return value should be `[32]byte` to match the underlying data type.

mcdee (changes requested):
Thank you for this; only one change requested.

pk910:
fixed :)

mcdee (approved):

PR #76, opened 2023-09-29
enable parsing of payload_attributes events

draganm:
Parse and pass `payload_attributes` events instead of ignoring them when received through SSE.


mcdee (approved):

mcdee:
Thank you for this contribution.

PR #77, opened 2023-09-30
[PoC] Implement response wrapper for ValidatorBalances

Willyham:
This is a proof of concept sketch for adding a generic response type which provides access to metadata. I intentionally picked an endpoint which had to make some decisions about how to handle chunking to get the ball rolling on discussion. There may be other questions to answer for some of the versioned endpoints as well.

### DIFF
--- a/api/v1/response.go
+++ b/api/v1/response.go
@@ -1,0 +1,28 @@
+package v1
+
+import "fmt"
+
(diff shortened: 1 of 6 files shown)


mcdee:
Thank you for your work on this; #80 has been merged so closing this item.

PR #80, opened 2023-10-06
Rework API

mcdee:
Rework of API to meet new requirements.  Main points:

- use of individual options struct for each endpoint
- use of `api.Response[T]` wrapper to provide additional information with response data

This is an alternative to #77 and has a wider scope.  It also tidies up various rough edges that have existed in the implementation.

@Willyham please take a look and let me know what you think in terms of the changes.  Note that these have not yet been applied across all APIs, that will be done prior to any merge but this should allow you to get the feel for the changes being made.


docs/0.19-changes.md
```diff
@@ -0,0 +1,93 @@
...
+```go
+validatorsProvider.Validators(ctx, "head", nil)
+```
+
+Even with a relatively simply call like this it is not clear what these parameters represent, especially when one is `nil`.  With the consolidation of call options this now looks something like this:
+
+```
+validatorsProvider.Validators(ctx, &api.ValidatorsOpts{
+  State: "head",
+})
```
Willyham, line 29: This is great 🙌

docs/0.19-changes.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Rationale behind changes
+
+The beacon API has changed significantly since it was first released.  These changes generally fall into three categories:
+
+- additional query parameters
+- additional response metadata in JSON and/or headers
+- introduction of new encoding systems, notably SSZ, for some endpoints
+
+Due to these changes, it has become increasingly difficult to provide full functionality without changing function signatures.  And, if function signatures are going to change, it makes sense to change them once in a more sustainable fashion rather than have a continual drip of function signature changes each time the beacon API changes.  0.19.0 brings these signature changes, and
```
Willyham, line 13: fyi dangling `and` here

docs/0.19-changes.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Response struct
+Responses from endpoints are now wrapped in the `api.Response` struct.  The requested information can be found in the `Data` element of this struct, and any metadata supplied in the response can be found in the `Metadata` element of this struct.  Metadata has been added to a number of endpoints to support situations where additional data may be useful to the user but is not part of the requested structure.  For example:
+
+```go
+response, _ := beaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
+  Block: "123",
+})
+
+finalized, exists := response.Metadata["finalized"]
```
Willyham, line 47: I do think it would be nice to provide constants for known metadata keys, seeing as they are defined in the spec. It would help avoid uncertainty around `executionOptimistic` vs `execution_optimistic` etc

docs/0.19-changes.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+Functions will now either return a response or an error, but always one and never both.  Combined with use of `api.Error` as mentioned above this allows for better control of both successful and failed calls, for example:
+
+```go
+response, err := beaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
+  Block: "999999999999",
+})
+if err != nil {
+  if apiErr, isAPiErr := err.(api.Error); isApiErr {
```
Willyham, line 76: Would be good to ensure this is compatible with `errors.As`

Willyham:
Left a few comments, I think this is definitely a good direction 👍

docs/0.19-changes.md
mcdee: Fixed; thanks.

docs/0.19-changes.md
mcdee: Added constants in `api/metadata`.

docs/0.19-changes.md
mcdee: Addressed and updated docs.

docs/0.19-changes.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+Functions will now either return a response or an error, but always one and never both.  Combined with use of `api.Error` as mentioned above this allows for better control of both successful and failed calls, for example:
+
+```go
+response, err := beaconBlockProvider.SignedBeaconBlock(ctx, &api.SignedBeaconBlockOpts{
+  Block: "999999999999",
+})
+if err != nil {
+var badInputErr *BadInputError
```
xenowits, line 76: this is not used in the code example below

mcdee: Fixed; thanks.

mcdee:
I'm relatively comfortable with the changes here, opening up for final review by interested parties prior to merging.

api/metadata/consts.go
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package metadata contains keys for well-known metadata fields provided in
+// an API response.
+package metadata
+
+const (
+	// Finalized is a boolean value describing if the response contains finalized data.
```
xenowits, line 19:
```suggestion
	// Finalized describes if the response contains finalized data.
```

xenowits: Similarly for `ExecutionOptimistic`

api/versionedproposal.go
```diff
@@ -0,0 +1,52 @@
...
+	Version   spec.DataVersion
+	Bellatrix *bellatrix.BeaconBlock
+	Capella   *capella.BeaconBlock
+	Deneb     *apiv1deneb.BlockContents
+}
+
+// String returns a string version of the structure.
+func (v *VersionedProposal) String() string {
+	switch v.Version {
+	case spec.DataVersionBellatrix:
```
xenowits, line 34: Why are we not checking for phase0 blocks?

api/versionedsignedproposal.go
```diff
@@ -0,0 +1,56 @@
...
+	Altair    *altair.SignedBeaconBlock
+	Bellatrix *bellatrix.SignedBeaconBlock
+	Capella   *capella.SignedBeaconBlock
+	Deneb     *apiv1deneb.SignedBlockContents
+}
+
+// String returns a string version of the structure.
+func (v *VersionedSignedProposal) String() string {
+	switch v.Version {
+	case spec.DataVersionBellatrix:
```
xenowits, line 38: Here as well, I believe we should add a switch case for phase0 version

xenowits (commented):
Overall, I think the changes are good to go!!

api/versionedproposal.go
mcdee: This was the result of a partial implementation during the move to using proposals.  Should now be fully updated with all forks.

PR #82, opened 2023-10-16
Basic Peering Provider

0xTylerHolmes:
Adds a basic interface and types to get a running nodes peers via the: /eth/v1/node/peers<?optional state&direction filters>


mcdee:
Thank you for this pull request.  The API has recently undergone a significant rework, so please can you update the code with the new style?  Thank you.

You should also create the equivalent functions and tests in the `mock` `multi` and `testclients` directories to make this function available for all common use cases and testing.

0xTylerHolmes:
Will do.

PR #83, opened 2023-10-19
add out-of-bounds check

xenowits:
Add an out-of-bounds check for `ValidatorState.String()`.

category: bug

### DIFF
--- a/api/v1/validatorstate.go
+++ b/api/v1/validatorstate.go
@@ -15,7 +15,6 @@ package v1
 
 import (
 	"fmt"
-	"github.com/pkg/errors"
 	"strings"
 
 	"github.com/attestantio/go-eth2-client/spec/phase0"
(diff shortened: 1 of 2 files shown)


api/v1/validatorstate.go
```diff
@@ -94,8 +94,12 @@ func (v *ValidatorState) UnmarshalJSON(input []byte) error {
 	return err
 }
 
-func (v ValidatorState) String() string {
-	return validatorStateStrings[v]
+func (v ValidatorState) String() (string, error) {
```
mcdee, line 97: This changes the function signature so it no longer implements stringer.  The signature should stay just returning string, and return something like "unknown state" if v is out of bounds.

api/v1/validatorstate.go
```diff
@@ -94,8 +94,12 @@ func (v *ValidatorState) UnmarshalJSON(input []byte) error {
...
+	if int(v) < 0 || int(v) >= len(validatorStateStrings) {
```
mcdee, line 98: The first recast isn't necessary; the compiler will convert the hardcoded 0 to a ValidatorState.

mcdee (changes requested):

api/v1/validatorstate.go
xenowits: yeah, good point. Updated to return `unknown` if v is out of bounds

mcdee (approved):

PR #84, opened 2023-10-21
Node peering rework

0xTylerHolmes:
NodePeer implementation with api rework, mock, test, and multi client implementations.

### DIFF
--- a/api/peeropts.go
+++ b/api/peeropts.go
@@ -1,0 +1,22 @@
+// Copyright © 2023 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// PeerOpts are the options for client side peer filtering.
+type PeerOpts struct {
+	// State of the connection (disconnected, connecting, connected, disconnecting)
+	State []string
+	// Direction of the connection (inbound, outbound)
+	Direction []string
+}
--- a/api/v1/attesterduty_test.go
+++ b/api/v1/attesterduty_test.go
@@ -18,8 +18,8 @@ import (
 	"testing"
 
 	api "github.com/attestantio/go-eth2-client/api/v1"
-	"github.com/stretchr/testify/assert"
 	require "github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestAttesterDutyJSON(t *testing.T) {
--- a/api/v1/attesterduty_test.go
+++ b/api/v1/attesterduty_test.go
@@ -18,8 +18,8 @@ import (
 	"testing"
 
 	api "github.com/attestantio/go-eth2-client/api/v1"
-	require "github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/assert"
+	require "github.com/stretchr/testify/require"
 )
 
 func TestAttesterDutyJSON(t *testing.T) {
--- a/api/v1/peers.go
+++ b/api/v1/peers.go
@@ -6,15 +6,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type Peers struct {
-	Peers []Peer
-}
-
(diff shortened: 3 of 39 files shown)


0xTylerHolmes:
see #82

api/v1/peers.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/pkg/errors"
+)
+
+type Peers struct {
+	Peers []Peer
+}
+
+type peersJSON struct {
+	Data []Peer                 `json:"data"`
+	Meta map[string]interface{} `json:"meta,omitempty"`
```
mcdee, line 16: I think that we should not support this field.  It isn't properly supported across beacon nodes, and the only information it provides can be calculated very easily.

http/nodepeers.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if len(additionalFields) > 0 {
+		request += "?"
+		for _, field := range additionalFields {
+			request += fmt.Sprintf("%s&", field)
+		}
+	}
+	//remove the trailing '&'
+	request = strings.TrimSuffix(request, "&")
```
mcdee: This is quite involved; would be easier to do a single `request = fmt.Sprintf("%s?%s", request, strings.Join(additionalFields,"&"))`

api/peeropts.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// PeerOpts are the options for client side peer filtering.
+type PeerOpts struct {
+	// State of the connection (disconnected, connecting, connected, disconnecting)
+	State []string
```
mcdee, line 19: This and `Direction` would be better off as enums (as much as we can make them).

api/v1/attesterduty_test.go
```diff
@@ -18,8 +18,8 @@ import (
 	"testing"
 
 	api "github.com/attestantio/go-eth2-client/api/v1"
-	require "github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/assert"
+	require "github.com/stretchr/testify/require"
```
mcdee, line 22: There seem to be a number of items in here that shouldn't be part of the PR; please can you remove them?

api/v1/peers.go
mcdee: This would also allow use to return `[]*Peer` rather than the `Peers` struct, which would be more in keeping with the rest of the API.

http/nodepeers.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, directionFilter := range opts.Direction {
+		additionalFields += fmt.Sprintf("direction=%s&", directionFilter)
+	}
+
+	if len(additionalFields) > 0 {
+		request += "?" + additionalFields
+		// remove the trailing '&'
+		request = strings.TrimSuffix(request, "&")
+	}
```
mcdee, line 36: I would make `additionalFields` a `[]string` and add each query parameter to the array individually.  Then you can easily combine them with a `fmt.Sprintf("%s?%s",request,strings.Join(additionalFields,"&"))`

This reduces allocations from repeatedly appending to the string.

mcdee (changes requested):
Thanks for the rework.  A few areas where either the underlying API is inconsistent or the return data is a little involved, apart from that looking good.

http/nodepeers.go
0xTylerHolmes: Ah interesting I reverted that as it felt clunky to make a list, iterate through the opts and then at the end check if the list was nonempty then append the strings. Ill revert to what I had if you prefer.

api/v1/attesterduty_test.go
0xTylerHolmes: Ah yes sorry I'm not sure how those made it in via the lint.

api/v1/peers.go
0xTylerHolmes: > This would also allow use to return `[]*Peer` rather than the `Peers` struct, which would be more in keeping with the rest of the API.

I like this better, I had been using round trip parsing of responses for the tests so I added it to conform. As its one way I think its best to just remove it.

api/peeropts.go
0xTylerHolmes: To do this would break the peeropts(api) / peer(v1) boundary. We can slide the peeropts into the v1 but that kind of sidesteps my interpretation of the refactor.

mcdee: We can leave it for now.

mcdee (commented):
Looks good; please could you address the few outstanding items from the linter, then should be good to go.

mcdee (approved):

PR #86, opened 2023-10-24
add Slot() to VersionedSignedProposal

xenowits:
Add `Slot()` method to `VersionedSignedProposal`.

category: feature

### DIFF
--- a/api/versionedsignedproposal.go
+++ b/api/versionedsignedproposal.go
@@ -14,7 +14,8 @@
 package api
 
 import (
-	apiv1deneb "github.com/attestantio/go-eth2-client/api/v1/deneb"
+	"errors"
+	"github.com/attestantio/go-eth2-client/api/v1/deneb"
 	"github.com/attestantio/go-eth2-client/spec"
 	"github.com/attestantio/go-eth2-client/spec/altair"
 	"github.com/attestantio/go-eth2-client/spec/bellatrix"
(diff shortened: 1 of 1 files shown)


api/versionedsignedproposal.go
```diff
@@ -14,7 +14,8 @@
 package api
 
 import (
-	apiv1deneb "github.com/attestantio/go-eth2-client/api/v1/deneb"
+	"errors"
+	"github.com/attestantio/go-eth2-client/api/v1/deneb"
```
mcdee, line 18: Please can we keep this import as `apiv1deneb`?  This helps to differentiate spec structs from those in the API.

api/versionedsignedproposal.go
```diff
@@ -29,7 +30,43 @@ type VersionedSignedProposal struct {
...
+	Deneb     *deneb.SignedBlockContents
+}
+
+// Slot returns the slot of the signed proposal.
+func Slot(p VersionedSignedProposal) (phase0.Slot, error) {
+	switch p.Version {
+	case spec.DataVersionPhase0:
+		if p.Phase0 == nil || p.Phase0.Message == nil {
+			return 0, errors.New("no phase0 block")
+		}
```
mcdee, line 42: If the code for the case is present then you may as well return the slot number, even though it's unlikely to ever be hit.  Or you can remove the case entirely, I don't mind which.

api/versionedsignedproposal.go
```diff
@@ -29,7 +30,43 @@ type VersionedSignedProposal struct {
...
+		if p.Deneb == nil || p.Deneb.SignedBlock == nil || p.Deneb.SignedBlock.Message == nil {
+			return 0, errors.New("no deneb block")
+		}
+
+		return p.Deneb.SignedBlock.Message.Slot, nil
+	default:
+		return 0, errors.New("unsupported version")
+	}
+
+	return 0, errors.New("unsupported version")
```
mcdee, line 69: This will not be necessary if the switch is exhaustive as per earlier comment.

mcdee (changes requested):
A couple of comments plus please lint the PR.  Thanks.

api/versionedsignedproposal.go
xenowits: yes i agree, phase0 code is very unlikely to get a hit. Removing the switch case for `DataVersionPhase0`

xenowits:
@mcdee I have resolved the review comments, could you please review again? 🙏

mcdee (approved):

Issue #87, opened 2023-10-24
Some spec values are not checked prior to start


0xTylerHolmes:
I'm adding this tool into some testnet testing infrastructure which uses different presets (minimal) and some custom config values. Due to this some of the assumed parameters used in parsing some of the objects are not valid. We should check the spec of the node and update these parameters prior to using endpoints which parse the beacon state.

I will submit a PR that I am currently using to work around this, but I am not sure if this is the best route to accomplish this.

mcdee:
There are significant issues with working outside the mainnet specification, not least with SSZ.  Different sized structures will result in different hash tree roots, and there will be issues with encoding to and decoding from SSZ to spec structs when sizes do not match the hard-coded values within the SSZ auto-generated code.

In general, the items that this module can support as variable are picked up when the service starts.  There may be one or two missing, but it isn't possible to run against an entirely different configuration such as minimal due to the aforementioned constraints.

0xTylerHolmes:
> There are significant issues with working outside the mainnet specification, not least with SSZ. Different sized structures will result in different hash tree roots, and there will be issues with encoding to and decoding from SSZ to spec structs when sizes do not match the hard-coded values within the SSZ auto-generated code.

Yeah it just hit me that all of the ssz generated objects will fail. I was enforcing JSON which allows all clients except for prysm to work fine. I'm fine just leaving my branch with this change if its going to be a headache. I can add a check that if the spec doesn't have the expected values for ssz then fail/warn if the user isn't enforcing JSON so no one runs into this issue down the line.

PR #88, opened 2023-10-24
set sync_committee_size on start

0xTylerHolmes:
A potential fix for #87. I figured hooking in at New would be sufficient. The only issue I can think of would be that having a multi with different testnet configs could be an issue. (though I am not sure why someone would do this) I am open to other options.

### DIFF
--- a/http/service.go
+++ b/http/service.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


0xTylerHolmes:
NOTE incomplete as is. Would require enforcing JSON when we detect that clients are running on a non-standard network

PR #93, opened 2023-11-10
Remove unused containers for deneb builder flow

avalonche:
remove containers deleted from https://github.com/ethereum/builder-specs/pull/90


api/versionedsignedblindedproposal.go
mcdee, line 1: I would like to retain this.  We should separate proposals from blocks, as it is likely that in future there will be other items that will need to be submitted with blocks as a proposal.

http/submitblindedproposal.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 // SubmitBlindedProposal submits a blinded proposal.
-func (s *Service) SubmitBlindedProposal(ctx context.Context, proposal *api.VersionedSignedBlindedProposal) error {
+func (s *Service) SubmitBlindedProposal(ctx context.Context, proposal *api.VersionedSignedBlindedBeaconBlock) error {
```
mcdee, line 28: Keep this as a proposal, as per previous comment.

service.go
mcdee, line 1: Keep this as proposal, as per previous comment.

mcdee (changes requested):

mcdee (approved):

PR #102, opened 2024-01-28
Pass `CommonOpts` downstream in `Validators`

moshe-blox:
For your consideration. Personally I'm not sure if this is right, perhaps we do prefer to fallback to default timeout when deviating from the user's call? (`Validators` -> `BeaconState`)


mcdee:
I have spent some time considering, and I think that it's valid to pass this along.

(At some stage we will be able to revert to using the validators endpoint, but that requires that endpoint to provide SSZ encoding, which is something I'm on the hook for.  It's on the list.)

mcdee (approved):

PR #107, opened 2024-02-25
feat: implement `produceBlockV3`

y0sher:
## Description

Implements [produceBlockV3](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV3) according to the ethereum beacon spec.

### Why

[produceBlockV2](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV2) is deprecated according to the ethereum beacon specs. the only correct way to get full blocks right now is using `produceBlockV3`. 

### Open questions 
- Should we leave the deprecated `produceBlockV2` to support legacy api endpoint? (I did for now)
- How to name the new V3 endpoint? I used `V3Proposal` for now, but I thought it's best to consult with you. (prysm for example called it `GenericVersionedBlock`

@mcdee your review appreciated


pinebit:
It turned out we worked simultaneously on the same problem :)
Still, I pushed my PR https://github.com/attestantio/go-eth2-client/pull/108 to show an alternative solution.
Feel free to discard if you want to proceed with yours. But we need this to be merged as soon as possible, as it blocks our work on [Charon](https://github.com/ObolNetwork/charon).

api/v3proposalopts.go
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import "github.com/attestantio/go-eth2-client/spec/phase0"
+
+// V3ProposalOpts are the options for obtaining v3 proposals.
+type V3ProposalOpts struct {
```
pinebit, line 19: This seems missing `builder_boost_factor` param.

api/versionedv3proposal.go
```diff
@@ -0,0 +1,870 @@
...
+	"github.com/attestantio/go-eth2-client/spec/bellatrix"
+	"github.com/attestantio/go-eth2-client/spec/capella"
+	"github.com/attestantio/go-eth2-client/spec/deneb"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+)
+
+// VersionedV3Proposal contains a versioned generic beacon block.
+type VersionedV3Proposal struct {
+	Version                 spec.DataVersion
+	ExecutionPayloadBlinded bool
```
pinebit, line 31: These should be automatically reflected in Metadata. Maybe we don't need to make them first-class citizens?

y0sher: Version is also in metadata but it is showed up in the struct, I took the example of how @mcdee did it with version. anyway lets wait for his review.

api/v3proposalopts.go
y0sher: thanks, it was actually added yesterday to the spec https://github.com/ethereum/beacon-APIs/pull/431, after I did the PR. will do the changes soon.

mcdee:
Hi thanks for the PR.  Internally we have been working on unifying the underlying blinded and non-blinded proposal flows, and have an alternate pr #116

Please could you take a look at this and let me know if it meets your requirements?  Thank you.

PR #111, opened 2024-02-28
Add nil checks to Versioned*BeaconBlock functions

jtraglia:
This PR makes these changes:

* Add a lot of missing nil checks in the `Versioned*BeaconBlock` files.
* Change "no deneb block contents" to "no deneb block" which I think are copy/paste mistakes.
* Fix a couple "denb" typos.
* Handle `DataVersionCapella` case in `BlobKZGCommitments` function.
* Remove `v.Capella.Message.Body.ExecutionPayload` check in `Grafitti` function.

There are two competing style for these nil checks. I used whichever was more prevalent in the file. Personally I like the split conditionals better, but it doesn't really matter. If you'd like me to use the same style, I'd be happy to make the change.

I'm made this PR without reaching out first because I'm under the assumption these will never actually be nil.

### DIFF
--- a/spec/versionedbeaconblock.go
+++ b/spec/versionedbeaconblock.go
@@ -67,7 +67,7 @@ func (v *VersionedBeaconBlock) Slot() (phase0.Slot, error) {
 		return v.Capella.Slot, nil
 	case DataVersionDeneb:
 		if v.Deneb == nil {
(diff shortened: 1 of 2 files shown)


mcdee (approved):
Looks good, thanks!

PR #116, opened 2024-03-02
Unified proposals using v3.

mcdee:
With the advent of the v3 blocks endpoint, it is possible to retrieve either a proposal or blinded proposal from the same endpoint.  Unify the underlying data structure into a single proposal, and support the v3 endpoint.

Also move to using new-style options for the related `Submit*` calls.


pinebit:
Is there an ETA for this to be merged and a new version to be released? Thanks!

mcdee:
> Is there an ETA for this to be merged and a new version to be released? Thanks!

There's an outstanding issue with one beacon node that I'd like to resolve prior to merging this (https://github.com/status-im/nimbus-eth2/issues/5850), and I wouldn't release until after the Deneb hard fork as pre-Deneb consensus nodes won't necessarily have support for this endpoint.

mcdee:
@pinebit supporting backward-compatibility can become very painful, and is not something that I'm comfortable with being able to provide.

My plan is to wait for all major clients (lighthouse,prysm,nimbus,teku) to support this endpoint to a given degree and then provide a release that uses this endpoint by preference.  Users of this repo can stick with the current release (0.19.10) until they are ready to upgrade their own codebases to use the v3 endpoint.

PR #121, opened 2024-03-25
Add an option for reduced memory usage

nkryuchkov:
This PR supersedes https://github.com/attestantio/go-eth2-client/pull/105 and uses the approach described in https://github.com/attestantio/go-eth2-client/pull/105#issuecomment-1937767726.

Changes:
- Add an option for HTTP Service that reduces memory usage by disabling certain actions that may take a significant amount of memory
- Optionally disable fetching validators from the state in the `Validators` call. The superseded PR has more details about this case

### DIFF
--- a/http/validators.go
+++ b/http/validators.go
@@ -119,12 +119,12 @@ func (s *Service) Validators(ctx context.Context,
 		return nil, errors.Join(errors.New("cannot specify both indices and public keys"), client.ErrInvalidOptions)
 	}
 
-	if !s.reducedMemoryUsage {
-		if len(opts.Indices) == 0 && len(opts.PubKeys) == 0 {
-			// Request is for all validators; fetch from state.
-			return s.validatorsFromState(ctx, opts)
-		}
+	if len(opts.Indices) == 0 && len(opts.PubKeys) == 0 {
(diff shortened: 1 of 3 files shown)


http/validators.go
```diff
@@ -119,14 +119,16 @@ func (s *Service) Validators(ctx context.Context,
 		return nil, errors.Join(errors.New("cannot specify both indices and public keys"), client.ErrInvalidOptions)
 	}
 
-	if len(opts.Indices) == 0 && len(opts.PubKeys) == 0 {
-		// Request is for all validators; fetch from state.
-		return s.validatorsFromState(ctx, opts)
-	}
+	if !s.reducedMemoryUsage {
+		if len(opts.Indices) == 0 && len(opts.PubKeys) == 0 {
```
mcdee, line 123: This part (request is for all validators) should be outside of the condition check; it is more efficient to obtain validators from state if all are requested, as we can use SSZ to pull the state so this gives us a smaller memory footprint than fetching just the validators with JSON.

mcdee (changes requested):
Thanks for the PR.  One change, apart from that it looks good.

mcdee (approved):

PR #122, opened 2024-03-25
`(g *Gwei) UnmarshalJSON()` `panic` on malformed input

infosecual:
I am currently running [Nosy](https://github.com/infosecual/nosy) on a broad set of ethereum client dependencies. It found a few panics while fuzzing `go-eth2-client`.  This fixes an issues in `(g *Gwei) UnmarshalJSON()` where a `panic: runtime error: slice bounds out of range` is triggerable due to a malformed JSON token. 

Summary:

`(g *Gwei) UnmarshalJSON()` has a few checks upon entry to prevent issues but a `panic` can still be triggered if a single `"` character is provided. 

```go
	if len(input) == 0 {
		return errors.New("input missing")
	}
	if !bytes.HasPrefix(input, []byte{'"'}) {
		return errors.New("invalid prefix")
	}
	if !bytes.HasSuffix(input, []byte{'"'}) {
		return errors.New("invalid suffix")
	}
```
This bypasses the `len(input) == 0` check as well as the prefix and suffix checks, since the single `"` character matches.

The panic happens on the slice indexing shortly after:
```go
	val, err := strconv.ParseUint(string(input[1:len(input)-1]), 10, 64)
```

This PR adds a minimum length check for a valid entry to `(g *Gwei) UnmarshalJSON()` as well as a test for the malformed input.

PR #125, opened 2024-04-01
`(s *Slot) UnmarshalJSON()` `panic` on malformed input

infosecual:
This PR is very similar to #122. This is a PR to fix a panic in a json unmarshal routine that was found via fuzzing. This PR adds a minimum length check for a valid entry to `(s *Slot) UnmarshalJSON()` as well as a test for the malformed input.


infosecual:
There are a few other instances of this same issue so I am closing this PR and just making one to fix them all

PR #126, opened 2024-04-01
Fix multiple `UnmarshalJSON()` `panics`

infosecual:
This is similar to #122. These panics where found while fuzzing with [Nosy](https://github.com/infosecual/nosy). 

This PR fixes the identified panics in:
- `(p *ParticipationFlags) UnmarshalJSON()`
- `(w *WithdrawalIndex) UnmarshalJSON()`
- `(b *BlobIndex) UnmarshalJSON()`
- `(s *Slot) UnmarshalJSON()`
- `(v *ValidatorIndex) UnmarshalJSON()`


mcdee:
The fixes look good, but the test filenames seem to be named somewhat oddly.  Please could you name them so that they match the filenames of the functions they are testing, for example putting `TestWithdrawalIndexUnmarshalJSON` in `withdrawalindex_test.go`?  Thanks.

infosecual:
Done. I split gwei_test.go into 3 separate files and renamed to correspond to the file that they test.

PR #128, opened 2024-04-05
Add a Makefile with a fmt rule

jtraglia:
This PR adds the ability to run `make fmt` which will automatically format everything. I think this would be a nice addition because (1) consistency is good and (2) after making a lot of changes it's nice knowing that everything is formatted properly and the imports are in a rational order.


mcdee:
I'm not a big fan of having additional commands that need to be run outside of the existing workflow, because it's easy for them to not be run.  That said, there is value in being stricter about formatting and the like.

I have added some additional linting rules and tidied up the code according to said rules and that seems to get us most of the way there.  This is easier to maintain, as it will run on every PR as part of the merge checks and allows them to be picked up by the PR author without us having to wonder if they ran the linting commands or not.

PR #136, opened 2024-05-07
Fix problems with new Electra structs

jtraglia:
This PR fixes the issues discovered by testing with the Electra reference structures.

* Add `electra.AggregationAndProof` structure.
* Add `electra.SignedAggregationAndProof` structure.
* Add `electra.AttesterSlashing` structure.
* Add `VersionedAttesterSlashing` structure.
* Add `VersionedIndexedAttestation` structure.
* Fix ssz-size for `CommitteeBits` (it's 8 bytes not 1 bit vector, confusing).
* Fix typo when unmarshaling consolidations.
* In `SignedConsolidation`, make message a pointer.
* Remove `ExitsRoot` from `ExecutionPayloadHeader`.
* Use yaml (not json) struct tag for `Consolidations`.
* Fix input to `MerkleizeWithMixin` for `ExecutionPayload`.

One of the tests is still failing, but I'm pretty sure the ref test is wrong here.

I sent this message to hww:

image


spec/electra/executionpayload_ssz.go
```diff
@@ -436,7 +436,7 @@ func (e *ExecutionPayload) HashTreeRootWith(hh ssz.HashWalker) (err error) {
 				hh.MerkleizeWithMixin(elemIndx, byteLen, (1073741824+31)/32)
 			}
 		}
-		hh.MerkleizeWithMixin(subIndx, num, 1073741824)
+		hh.MerkleizeWithMixin(subIndx, num, 1048576)
```
jtraglia, line 439: It took me quite a while to find this issue. Why was this auto-generated value used? Have you been manually changing this for previous versions?

mcdee: This was addressed in sszgen as per https://github.com/ferranbt/fastssz/issues/127 but not sure if it has made it into a release version yet.

mcdee:
Extra data of `0x` is an empty value.  In YAML strings don't have to be quoted, as far as I'm aware.  You will find that consensus and execution nodes are a little lax when it comes to formatting, so it's likely that you will have to handle that situation.

jtraglia:
@mcdee I'm not entirely sure how to fix this `0x` issue. I can fix our marshaled output, but the test code will marshal it to a string, which has a different value. It will marshal it to `0` (not `0x`) there.

```go
func testYAMLFormat(input []byte) string {
	val := make(map[string]any)
	if err := yaml.UnmarshalWithOptions(input, &val, yaml.UseOrderedMap()); err != nil {
		panic(err)
	}
        ...
```

For example, the following. The top line is the expected output, the bottom line is the actual output:

image


Any advice here?

jtraglia:
Ah I found that you did this for Capella. I'll copy this:

image

jtraglia:
Pushed a commit which handles that edge case. Also, Hsiao-Wei has acknowledged the issue. I'm going to make a PR to the consensus-specs to get that fixed. The fix might not make it into the next release though.

mcdee (approved):

PR #140, opened 2024-05-14
Electra minimal preset fixes & add Consolidations getter

pk910:
* Fixed use of dynamic ssz for electra structs when minimal spec support enabled
* Added generic getter for Consolidations


pk910:
fixed a bug in `dynamic-ssz` too and bumped version to v0.0.4.
works fine for dora with minimal preset on electra now :)

mcdee (approved):

PR #162, opened 2024-09-20
Add beacon block reward endpoint

MisiakGeo:
Add Block Rewards. 
Endpoint `/eth/v1/beacon/rewards/blocks` is now available to fetch data regarding the block rewards on a slot.

To test it you can run the following:

```go
package main

import (
	"context"
	"errors"
	"fmt"
	"os"
	"time"

	"github.com/attestantio/go-eth2-client/api"
	"github.com/attestantio/go-eth2-client/http"
	"github.com/davecgh/go-spew/spew"
	"github.com/rs/zerolog"
)

func main() {
	// Provide a cancellable context to the creation function.
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	var err error
	client, err := http.New(ctx,
		// WithAddress supplies the address of the beacon node, as a URL.
		http.WithAddress(os.Getenv("BEACON_NODE_ADDRESS")),
		// LogLevel supplies the level of logging to carry out.
		http.WithLogLevel(zerolog.Disabled),
	)
	if err != nil {
		panic(err)
	}

	fmt.Printf("Connected to %s\n", client.Name())

	httpClient := client.(*http.Service)
	reward, err := httpClient.BeaconBlockReward(ctx, &api.BeaconBlockRewardOpts{
		Common: api.CommonOpts{
			Timeout: 2 * time.Minute,
		},
		Block: "head",
	})
	if err != nil {
		var apiErr *api.Error
		if errors.As(err, &apiErr) {
			switch apiErr.StatusCode {
			case 400:
				// Bad Request
			case 404:
				// No block found.
			case 503:
				// Node is syncing.
			case 500:
				// Internal Server Error.
			}
		}
		spew.Dump(err)
		panic(err)
	}
	spew.Dump(reward.Data)

	// Cancelling the context passed to New() frees up resources held by the
	// client, closes connections, clears handlers, etc.
	cancel()
}
```

One problem I had, was when testing it against QuickNode, some time an error appeared `Context Deadline Exceeded`, due to the fact that the rewards endpoint with curl takes about 1 min to complete.

### DIFF
--- a/api/v1/beaconblockreward.go
+++ b/api/v1/beaconblockreward.go
@@ -1,0 +1,146 @@
+// Copyright © 2020, 2021 Attestant Limited.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+// BeaconBlockReward Rewards info for a single block
+type BeaconBlockReward struct {
+	// Proposer of the block, the proposer index who receives these rewards
+	ProposerIndex uint64
+	// Total block reward in gwei, equal to attestations + sync_aggregate + proposer_slashings + attester_slashings
+	Total uint64
+	// Block reward component due to included attestations in gwei
+	Attestations uint64
+	// Block reward component due to included sync_aggregate in gwei
+	SyncAggregate uint64
+	// Block reward component due to included proposer_slashings in gwei
+	ProposerSlashings uint64
+	// Block reward component due to included attester_slashings in gwei
+	AttesterSlashings uint64
+}
+
+// beaconBlockRewardJSON is the spec representation of the struct.
+type beaconBlockRewardJSON struct {
+	ProposerIndex     string `json:"proposer_index"`
+	Total             string `json:"total"`
+	Attestations      string `json:"attestations"`
+	SyncAggregate     string `json:"sync_aggregate"`
+	ProposerSlashings string `json:"proposer_slashings"`
+	AttesterSlashings string `json:"attester_slashings"`
+}
+
+// MarshalJSON implements json.Marshaler.
+func (b *BeaconBlockReward) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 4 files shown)


MisiakGeo:
Ignore the events.go (It was fixed by @mcdee). It is still working perfectly without errors, but `urlForCall` function is considered better.

http/beaconblockreward.go
```diff
@@ -0,0 +1,55 @@
...
+	"bytes"
+	"context"
+	"fmt"
+
+	client "github.com/attestantio/go-eth2-client"
+	"github.com/attestantio/go-eth2-client/api"
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+)
+
+// BeaconBlockHeader provides the block header given the opts.
```
mcdee, line 26: Please could you fix this comment to match the function.

api/v1/beaconblockreward.go
```diff
@@ -0,0 +1,146 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+// BeaconBlockReward Rewards info for a single block
```
mcdee, line 24: Please can you fix comments so that they end in a period?  Thanks.

api/v1/beaconblockreward.go
```diff
@@ -0,0 +1,146 @@
...
+type BeaconBlockReward struct {
+	// Proposer of the block, the proposer index who receives these rewards
+	ProposerIndex uint64
```
mcdee, line 27: This should be of type `phase0.ValidatorIndex`

api/v1/beaconblockreward.go
```diff
@@ -0,0 +1,146 @@
...
+	// Total block reward in gwei, equal to attestations + sync_aggregate + proposer_slashings + attester_slashings
+	Total uint64
```
mcdee, line 29: This and other gwei values should be of type `phase0.Gwei`

http/events.go
```diff
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"net/url"
```
mcdee, line 24: Please can you revert the changes to this file, as they have been addressed elsewhere.

mcdee (changes requested):
Thanks for this PR, a few comments inline.  Also, please could you add in `api/v1/beaconblockreward_yaml.go` and `api/v1/beaconblockreward_test.go` files to add YAML support and tests for the spec, respectively?  The tests in particular are important to keep up coverage rates for the library.

MisiakGeo:
@mcdee Thank you very for the comments. I will take a look. A comment from my side. YAML files can be found in `api/v1/daneb/*_yaml.go` . Should I place them in `api/v1`?

mcdee:
Sorry, after taking another look it appears that the rewards object is not a spec object, so it doesn't need the YAML piece.

mcdee:
Superseded by 8910181b413ea0d482141be8d5440cd01c37349f

PR #172, opened 2024-11-04
Add SingleAttestation container

Bez625:
As per https://github.com/ethereum/consensus-specs/pull/3900 - introduce a SingleAttestation container


spec/electra/singleattestation_json.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return errors.Wrap(err, "invalid JSON")
+	}
+
+	return a.unpack(&singleAttestationJSON)
+}
+
+func (a *SingleAttestation) unpack(singleAttestationJSON *singleAttestationJSON) error {
+	var err error
+	if singleAttestationJSON.CommitteeIndex == "" {
```
mcdee, line 57: It doesn't look like committee index or attester index is actually populated.  We should have at the least a test for a correct JSON version of the attestation, and ideally the standard incorrect ones as well.  The phase0 tests for attestation can be used as a basis.

mcdee (changes requested):

mcdee (approved):

PR #182, opened 2025-01-16
Update aggregate attestation fetching and submitting for Electra

Bez625:
Adding versioned attestation aggregate fetching and submitting

Compliment to https://github.com/attestantio/vouch/pull/284

PR #183, opened 2025-01-21
Add signature utility functions

Bez625:
Utility methods requested in this comment https://github.com/attestantio/go-eth2-client/pull/182#issuecomment-2605729905

### DIFF
--- a/spec/versionedaggregateandproof.go
+++ b/spec/versionedaggregateandproof.go
@@ -123,3 +123,47 @@ func (v *VersionedAggregateAndProof) String() string {
 		return "unknown version"
 	}
 }
+
+// SelectionProof returns the selection proof of the aggregate.
+func (v *VersionedAggregateAndProof) SelectionProof() (phase0.BLSSignature, error) {
+	switch v.Version {
+	case DataVersionPhase0:
+		if v.Phase0 == nil {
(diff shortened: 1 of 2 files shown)


mcdee (approved):

KaloyanTanev:
Much appreciated!

KaloyanTanev:
@Bez625 do you think adding `AggregatorIndex` as well makes sense?

Bez625:
> @Bez625 do you think adding `AggregatorIndex()` as well makes sense? And also adding `SelectionProof()` to the Versioned counterpart?

It's definitely possible to add those, but I can't see anywhere we would use them at the minute. Are they functions you would find useful? If so I'm happy to raise the change

KaloyanTanev:
Yes, we have a couple of places throughout our Charon codebase where those will be quite handy. One example is [here](https://github.com/ObolNetwork/charon/blob/e8f02f63a9e20ac539b0fbaeb4b0e2c4ec8b1448/core/validatorapi/validatorapi.go#L796). I am currently refactoring the code to encompass the `SubmitAggregateAttestationsOpts` (and subsequently `VersionedSignedAggregateAndProof`) and being able to call directly `agg.AggregatorIndex()` as we've been doing previously with `agg.Message.AggregatorIndex` would be nice.

PR #185, opened 2025-01-25
chore: posted a link with a workflow

Lubov66:
found that this reference was omitted in the documentation


mcdee:
Thank you for your contribution, however we do not accept typo/spelling/grammar pull requests for security reasons.

PR #186, opened 2025-01-27
Update attestation and aggregate submission

Bez625:
- Correct query params for fetching aggregate attestation
- Updates to VersionedAttestations retrieval from beacon block
- Update verify on aggregate data based on electra

### DIFF
--- a/api/attestationdataopts.go
+++ b/api/attestationdataopts.go
@@ -14,7 +14,6 @@
 package api
 
 import (
-	"github.com/attestantio/go-eth2-client/spec"
 	"github.com/attestantio/go-eth2-client/spec/phase0"
 )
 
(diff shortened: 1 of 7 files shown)


api/attestationdataopts.go
```diff
@@ -23,4 +26,6 @@ type AttestationDataOpts struct {
 	Slot phase0.Slot
 	// CommitteeIndex is the committee index for which the data is obtained.
 	CommitteeIndex phase0.CommitteeIndex
+	// Version is the version of the chain we are using. Used to control fork specific logic.
+	Version spec.DataVersion
```
mcdee, line 30: Do we need this, given that we are passed the slot and the library knows the epochs of the hard fork boundaries?

mcdee (commented):
Question on the requirement of `Version` within the attestation data opts.

api/attestationdataopts.go
Bez625: I wasn't entirely sure how to do this so I've made an attempt to use the Spec to retrieve the electra epoch and convert. Please let me know if there is a better way to achieve this

mcdee (approved):

PR #190, opened 2025-01-31
Fix submissions to use SingleAttestation

Bez625:


http/submitattestations.go
```diff
@@ -95,7 +95,11 @@ func createUnversionedAttestations(attestations []*spec.VersionedAttestation) ([
 		case spec.DataVersionDeneb:
 			unversionedAttestations = append(unversionedAttestations, attestations[i].Deneb)
 		case spec.DataVersionElectra:
-			unversionedAttestations = append(unversionedAttestations, attestations[i].Electra)
+			singleAttestation, err := attestations[i].Electra.ToSingleAttestation(attestations[i].ValidatorIndex)
+			if err != nil {
+				return nil, errors.Join(errors.New("failed to convert attestation to single attestation"), err)
```
mcdee, line 100: In this situation I think we should log an error but continue.  The rationale behind this is that we want to submit as many attestations as possible, so if there is a failure for one of the potentially many attestations in this request we don't want to reject all of them.

mcdee (changes requested):

http/submitattestations.go
```diff
@@ -95,7 +95,14 @@ func createUnversionedAttestations(attestations []*spec.VersionedAttestation) ([
 		case spec.DataVersionDeneb:
 			unversionedAttestations = append(unversionedAttestations, attestations[i].Deneb)
 		case spec.DataVersionElectra:
-			unversionedAttestations = append(unversionedAttestations, attestations[i].Electra)
+			singleAttestation, err := attestations[i].Electra.ToSingleAttestation(attestations[i].ValidatorIndex)
+			if err != nil {
+				s.log.Warn().Err(errors.Join(errors.New("failed to convert attestation to single attestation"), err)).
```
mcdee, line 100: I don't think you need that Join here, as it contains the same message as you are logging.

mcdee (changes requested):

mcdee (approved):

Issue #171, opened 2024-10-29
/eth/v1/node/syncing endpoint does not return correct output as per beacon spec


nidhi-singh02:
The beacon API [spec](https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus) returns the output below

```
{
  "data": {
    "head_slot": "1",
    "sync_distance": "1",
    "is_syncing": true,
    "is_optimistic": true,
    "el_offline": true
  }
}
```

However the node syncing endpoint in here returns `*api.Response[*apiv1.SyncState]` and this is how SyncState looks like- it is missing field el_offline. 
```
// SyncState is the data regarding the node's synchronization state to the chain.
type SyncState struct {
	// HeadSlot is the head slot of the chain as understood by the node.
	HeadSlot phase0.Slot
	// SyncDistance is the distance between the node's highest synced slot and the head slot.
	SyncDistance phase0.Slot
	// IsOptimistic is true if the node is optimistic.
	IsOptimistic bool
	// IsSyncing is true if the node is syncing.
	IsSyncing bool
}

```

I am happy to raise a PR to fix this.

mcdee:
This was a relatively recent addition, and was missed.  A PR to address this would be appreciated.

PR #192, opened 2025-02-03
CheckConnectionState: option to consider node as not synced if el_offline is true

nkryuchkov:
Closes https://github.com/attestantio/go-eth2-client/issues/171

The `multi` client doesn't switch clients if the current client's EL goes offline until the client sets `is_syncing` to `true`. It causes duty misses.

The PR adds a `WithELConnectionCheck` parameter to both `http` and `multi`, which makes the client consider a node as synced only if `el_offline` is `false`

It seems to work with Teku, Lighthouse, Nimbus correctly. It doesn't work with Prysm because of https://github.com/prysmaticlabs/prysm/issues/14226, its implementation of `el_offline` is incorrect, so the switch won't happen until it starts returning `"is_syncing":true`.

### DIFF
--- a/api/v1/syncstate.go
+++ b/api/v1/syncstate.go
@@ -32,6 +32,8 @@ type SyncState struct {
 	IsOptimistic bool
 	// IsSyncing is true if the node is syncing.
 	IsSyncing bool
+	// ELOffline is true if the node's EL client is offline.
+	ELOffline bool
 }
 
 // syncStateJSON is the spec representation of the struct.
@@ -40,6 +42,7 @@ type syncStateJSON struct {
 	SyncDistance string `json:"sync_distance"`
 	IsOptimistic bool   `json:"is_optimistic"`
 	IsSyncing    bool   `json:"is_syncing"`
+	ELOffline    bool   `json:"el_offline"`
 }
 
 // MarshalJSON implements json.Marshaler.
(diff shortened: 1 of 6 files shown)


nkryuchkov:
Linter and tests seem to fail due to the same reasons as in https://github.com/attestantio/go-eth2-client/pull/191

mcdee:
Which beacon node(s) exhibit this behavior?  And is this an issue with the beacon node not supplying duties, or is it that the multi client fails to recognize that the beacon node is not supplying duties?

nkryuchkov:
@mcdee Sorry for the delay. We at @ssvlabs tested this solution for a while and noticed cases when the Prysm node had been mistakenly considered unhealthy because `SyncDistance` grew a little. So we decided not to use this solution because the Prysm workaround is too hacky, and increasing the allowed `SyncDistance` gap doesn't look good either. 

I removed the Prysm workaround, but unfortunately, https://github.com/prysmaticlabs/prysm/issues/14226 hasn't been fixed yet, so this PR wouldn't work for Prysm

> Which beacon node(s) exhibit this behavior? 

I tested it with Prysm, Lighthouse, Teku, Nimbus. All clients except Prysm set `el_offline` correctly, but Prysm didn't

> And is this an issue with the beacon node not supplying duties, or is it that the multi client fails to recognize that the beacon node is not supplying duties?

The multi-client fails to recognize that the beacon node is not supplying duties (the beacon node sets `el_offline` to `true` but is still synced)

PR #196, opened 2025-02-05
fix dynssz annotation for `electra.Attestation`

pk910:
Parsing electra blocks with dynssz enabled currently fails for the BeaconBlockBody.Attestations field (size mismatch).

The problem is a wrong annotation in the Attestation container.
`dynssz-size` shouldn't be set here as there is no `ssz-size` too.
The intention of that tag was probably to calculate the `ssz-max` tag dynamically. 
The corresponding dynssz tag is called `dynssz-max`.

### DIFF
--- a/spec/electra/attestation.go
+++ b/spec/electra/attestation.go
@@ -31,7 +31,7 @@ import (
 //
 //nolint:tagalign
 type Attestation struct {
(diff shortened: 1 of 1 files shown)


pk910:
@mcdee 
ready for review :)

mcdee (approved):

Issue #197, opened 2025-02-10
SSZ responses with no body are parsed as JSON


samcm:
I've been tracking down an issue we've observed in Pectra Devnet 6 where we are failing to parse blobs from some slots. These slots don't have any blobs.

The issue appears to be in https://github.com/attestantio/go-eth2-client/blob/f19fd9b017dc9b3dff6ae209f3b08e08cc644056/http/http.go#L403-L427

We fall back to `ContentTypeJSON` if the reponse body is empty, despite an accurate `Content-Type=application/octet-stream`existing in the response.

This throws out the content type parsing further up the chain trying to parse an empty body as JSON, resulting in a log/error like this: `failed to parse JSON\nEOF`

PR #198, opened 2025-02-10
Only fall back to ContentTypeJSON  when there is no `Content-Type` header & an empty response body

samcm:
Resolves https://github.com/attestantio/go-eth2-client/issues/197


mcdee:
At first glance this results in different responses to the `BlobSidecars()` call when requesting SSZ and JSON, if there are no blobs.  Requesting SSZ will return an empty array, and requesting JSON will error.  This doesn't seem like a great outcome, so let me look at the parsing code to see if we can provide a similar response for each.

samcm:
@mcdee Yup that makes perfect sense - definitely happy with that approach! Thanks for having a look at thie. Appreciate it.

mcdee (approved):

PR #200, opened 2025-02-11
feat: Add Validator Liveness support

olegshmuelov:
Add support for the POST `/eth/v1/validator/liveness/{epoch}` endpoint in the Beacon Node API.
This endpoint checks if a validator has been observed as live during a given epoch.


mcdee:
Looks good, thanks for adding this.

mcdee (approved):

PR #210, opened 2025-02-17
Refactor events handling for linter

Bez625:


Bez625:
Apologies, I switched the map for a case statement as I didn't want to create all the structs on every event.

PR #211, opened 2025-02-18
add `dynssz-max` tags

pk910:
This PR adds `dynssz-max` tags for all fields that have a `ssz-max` tag.
This does not affect any of the containers right now,  but is required for generating hash tree roots of containers for non-mainnet presets.

It'd help me a lot to add these annotations to the containers already.


pk910:
@mcdee ready for review, but logic wise nothing changed so far.

spec/electra/attestation.go
```diff
@@ -28,14 +28,11 @@ import (
...
-//nolint:tagalign
 type Attestation struct {
-	AggregationBits bitfield.Bitlist `ssz-max:"131072" dynssz-max:"MAX_VALIDATORS_PER_COMMITTEE*MAX_COMMITTEES_PER_SLOT"`
+	AggregationBits bitfield.Bitlist `dynssz-max:"MAX_VALIDATORS_PER_COMMITTEE*MAX_COMMITTEES_PER_SLOT" ssz-max:"131072"`
 	Data            *phase0.AttestationData
-	Signature       phase0.BLSSignature `ssz-size:"96"`
-	// bitfield.Bitvector64 is an 8 byte array so dynamic sizing doesn't make sense.
-	CommitteeBits bitfield.Bitvector64 `ssz-size:"8"`
+	Signature       phase0.BLSSignature  `ssz-size:"96"`
+	CommitteeBits   bitfield.Bitvector64 `dynssz-size:"MAX_COMMITTEES_PER_SLOT/8" ssz-size:"8"`
```
mcdee, line 35: As per the comment, this is not a dynamic array but a Bitvector64 so of fixed size.  What would the `dynssz-size` tag change here?

spec/electra/beaconblockbody.go
```diff
@@ -24,19 +24,21 @@ import (
...
 	ETH1Data              *phase0.ETH1Data
 	Graffiti              [32]byte                      `ssz-size:"32"`
-	ProposerSlashings     []*phase0.ProposerSlashing    `ssz-max:"16"`
-	AttesterSlashings     []*AttesterSlashing           `ssz-max:"1"`
-	Attestations          []*Attestation                `ssz-max:"8"`
-	Deposits              []*phase0.Deposit             `ssz-max:"16"`
-	VoluntaryExits        []*phase0.SignedVoluntaryExit `ssz-max:"16"`
+	ProposerSlashings     []*phase0.ProposerSlashing    `dynssz-max:"MAX_PROPOSER_SLASHINGS" ssz-max:"16"`
+	AttesterSlashings     []*AttesterSlashing           `dynssz-max:"MAX_ATTESTER_SLASHINGS" ssz-max:"1"`
+	Attestations          []*Attestation                `dynssz-max:"MAX_ATTESTATIONS"       ssz-max:"8"`
```
mcdee, line 35: Shouldn't this be `MAX_ATTESTATIONS_ELECTRA`?

spec/electra/beaconblockbody.go
```diff
@@ -24,19 +24,21 @@ import (
...
 	RANDAOReveal          phase0.BLSSignature `ssz-size:"96"`
 	ETH1Data              *phase0.ETH1Data
 	Graffiti              [32]byte                      `ssz-size:"32"`
-	ProposerSlashings     []*phase0.ProposerSlashing    `ssz-max:"16"`
-	AttesterSlashings     []*AttesterSlashing           `ssz-max:"1"`
-	Attestations          []*Attestation                `ssz-max:"8"`
-	Deposits              []*phase0.Deposit             `ssz-max:"16"`
-	VoluntaryExits        []*phase0.SignedVoluntaryExit `ssz-max:"16"`
+	ProposerSlashings     []*phase0.ProposerSlashing    `dynssz-max:"MAX_PROPOSER_SLASHINGS" ssz-max:"16"`
+	AttesterSlashings     []*AttesterSlashing           `dynssz-max:"MAX_ATTESTER_SLASHINGS" ssz-max:"1"`
```
mcdee, line 34: Shouldn't this be `MAX_ATTESTER_SLASHINGS_ELECTRA`?

mcdee (changes requested):
A few fixes required, I think.

spec/electra/attestation.go
pk910: Yea, I think the `bitfield.Bitvector64` type is not ideal here.
The underlying base type is still a dynamic byte slice (`[]byte`), so the generic marshaller will just fill it with the given amount of bytes. That works fine even for repeated marshalling / unmarshalling, as long as the methods of `Bitvector64` are not gonna used excessively.

I've checked the dora codebase and I actually applied a workaround by not using the `Bitvector64.BitAt` method which fails with non expected array length, but using the `Bitvector64.BitIndices` method which works fine :D

Still definitely not ideal. The tag needs to be there anyway as otherwise the dynamic marshalling using the minimal preset fails. There are simply not 4 bytes available in a minimal block ssz stream.

pk910: I've double checked, the removal of that tag broke the "weak support" for minimal preset blocks some time ago:
`failed to decode electra signed block contents\nfailed decoding field Message: failed decoding field Body: failed decoding field Attestations: incorrect size`

Works fine again with this branch :)   (we actually haven't used a minimal preset devnet for some time, so didn't really notice it yet)

spec/electra/beaconblockbody.go
pk910: that actually solved an issue that I was looking for since quite some time :D  thx :)

spec/electra/attestation.go
mcdee: Okay, I think we're stuck with keeping this as a `Bitvector64` for mainnet compatibility, but understand the requirement for minimal.

mcdee (approved):

PR #213, opened 2025-03-10
spec: Implement DataVersionFromString

MariusVanDerWijden:
Implements DataVersionFromString to turn a string into a DataVersion

### DIFF
--- a/spec/dataversion.go
+++ b/spec/dataversion.go
@@ -48,6 +48,16 @@ var dataVersionStrings = [...]string{
 	"electra",
 }
 
+var stringToDataVersion = map[string]DataVersion{
+	dataVersionStrings[0]: DataVersionUnknown,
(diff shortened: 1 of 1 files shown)


mcdee:
Thank you for this PR.  On review I think I would prefer it if `DataVersionFromString()` used `UnmarshalJSON()` rather than the other way around, _i.e._ `DataVersionFromString()` took the input, added quotes, and passed it to the existing `UnmarshalJSON()`.  This would have a couple of benefits:

- be case-insensitive to the input (historically we have seen code that passes around fork names in upper-case, for example)
- does not make an assumption about the outer characters of the input in `UnmarshalJSON()`

And it should also result in a much smaller PR.

mcdee:
Looks like a couple of linting errors, if you could fix those I think we're good to merge.  Thanks!

PR #215, opened 2025-03-14
Add pending deposits api

Bez625:


mcdee:
Which beacon nodes support this endpoint?  We are usually a little careful in releasing new endpoints until they have wider support.

mcdee (approved):

PR #216, opened 2025-03-15
chore: make function comment match function name

welfuture:
make function comment match function name


mcdee:
Thank you for your contribution, however we do not accept typo/spelling/grammar pull requests for security reasons.

PR #217, opened 2025-03-20
fix: add `dynssz-max` tag to capella withdrawals

filoozom:
This was already the case in capella's and deneb's `ExecutionPayload` structures, but missing here, unless I'm mistaken?


mcdee (approved):

mcdee:
Thanks for this fix.

PR #221, opened 2025-04-15
Update Consensus Block Value Handling

MatusKysel:
•	Consensus Block Value: Now only returns an error if a non-empty value fails conversion, aligning with [PR14111](https://github.com/OffchainLabs/prysm/pull/14111) that it isn’t critical for block production.
•	Other Headers: No changes; the execution and blinded fields continue to be validated as before.

This update improves robustness by preventing errors when the consensus value is absent.

### DIFF
--- a/http/proposal.go
+++ b/http/proposal.go
@@ -350,7 +350,9 @@ func (*Service) populateProposalDataFromHeaders(response *api.Response[*api.Vers
 		case strings.EqualFold(k, "Eth-Consensus-Block-Value"):
 			var success bool
 			response.Data.ConsensusValue, success = new(big.Int).SetString(v, 10)
-			if !success {
+			// https://github.com/OffchainLabs/prysm/pull/14111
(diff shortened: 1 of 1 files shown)


mcdee:
I'm not convinced by this change.  The spec requires the value to be present, so if it is not then that suggests an error with the server.  The assumption that the block value is not important to the code receiving the response is not valid in all situations, so this change could end up masking a real problem.

mcdee:
Although I understand the idea, I am concerned that letting through blocks that have an equivalent 0 value is going to cause problems itself by masking the error and providing incorrect information to clients.

More generally, patching around nodes that fail to adhere to the spec creates a precedent that I'm not sure we want to create.  We would be stuck with these workarounds in the codebase for months or years (most likely until the next hard fork), and it creates additional work to confirm that any fix is in place prior to us being able to remove the patch.  As such, I think the best we can realistically manage is to ensure we operate according to the spec.

PR #222, opened 2025-05-05
Update go-bitfield import path to OffchainLabs

estensen:
After Prysm team joined Offchain Labs they moved their repo into the Offchain Labs org.
https://github.com/OffchainLabs/go-bitfield/pull/63


Bez625:
@estensen thanks for the contribution, this was updated separately.

Issue #237, opened 2025-05-26
Make Electra execution request fields optional for beacon client compatibility


klim0v:
### Problem Statement

The current implementation strictly validates all Electra execution request fields (`deposit_requests`, `withdrawal_requests`, `consolidation_requests`), requiring them to be present and non-nil. However, some beacon chain clients omit these fields entirely when they contain no data, causing parsing failures.

This issue is also documented in: https://github.com/sigp/lighthouse/issues/7520

### Current Implementation
```go
if data.DepositRequests == nil {
    return errors.New("payload attributes deposit requests missing")
}
for i := range data.DepositRequests {
    if data.DepositRequests[i] == nil {
       return fmt.Errorf("deposit requests entry %d missing", i)
    }
}
p.DepositRequests = data.DepositRequests

if data.WithdrawalRequests == nil {
    return errors.New("payload attributes withdraw requests missing")
}
for i := range data.WithdrawalRequests {
    if data.WithdrawalRequests[i] == nil {
       return fmt.Errorf("withdraw requests entry %d missing", i)
    }
}
p.WithdrawalRequests = data.WithdrawalRequests

if data.ConsolidationRequests == nil {
    return errors.New("payload attributes consolidation requests missing")
}
for i := range data.ConsolidationRequests {
    if data.ConsolidationRequests[i] == nil {
       return fmt.Errorf("consolidation requests entry %d missing", i)
    }
}
p.ConsolidationRequests = data.ConsolidationRequests
```

### Rationale

**Client Implementation Differences:**
- **Lighthouse**: Omits execution request fields when they contain no data
- **Other clients**: May include empty arrays `[]` for consistency
- **Specification**: Defines fields as lists that can be empty, but doesn't mandate JSON representation when empty

**Why This Approach:**
1. **Maintains Data Integrity**: Still validates individual entries when present
2. **Improves Compatibility**: Works with all known beacon client implementations  
3. **Follows Specification**: Empty lists are semantically equivalent to omitted fields
4. **Backward Compatible**: Existing functionality remains unchanged

### Impact Assessment

**Benefits:**
- Resolves parsing failures with Lighthouse beacon client
- Maintains strict validation of actual data entries
- No breaking changes to existing API
- Improves library robustness across different client implementations

**Risks:**
- Minimal - only relaxes field presence requirements while preserving data validation

### References

- [Issue #236: PayloadAttributes parsing failures with Lighthouse](https://github.com/sigp/lighthouse/issues/7520)
[...]

PR #238, opened 2025-05-26
remove not nil check for payloadAttributes WithdrawalRequests and ConsolidationRequests

klim0v:
I'm also not sure if Prysm and other clients support these fields: https://github.com/attestantio/go-eth2-client/issues/237

https://github.com/OffchainLabs/prysm/blob/0a48fafc710bd4453d71fe44d298c9f24a763787/beacon-chain/rpc/eth/events/events.go#L806C6-L846


klim0v:
@Bez625 any thoughts here?

klim0v:
@mcdee, @pk910

Issue #227, opened 2025-05-19
Allow easy use of new EF Minimal specs.


ZanCorDX:
EF is starting to use (on devnet tests) minimal specs instead of the usual main net specs to speed up everything.
These changes impact on the ssz definitions (eg: https://github.com/attestantio/go-eth2-client/blob/master/spec/electra/executionrequests.go) that are used to compute the hash that is used to sign bids on mev-boost-relay (eg: https://github.com/flashbots/mev-boost-relay/blob/344d5b67d5940b2bd4e462d790858dd05a445fdc/common/types_spec.go#L188).
It would be great to have simple out of the box support for those minimal specs.

PR #241, opened 2025-06-03
feat: export minimal/mainnet presets

jacobkaufmann:
embed and export items for mainnet/minimal ethereum consensus specs presets.

this is related to #227.


jacobkaufmann:
while it is not ideal to maintain separate copies of the consensus specs presets, the presets should not change except for the fork under development.

the git submodule approach failed because the `go:embed` files were not available to an importer of the package.

PR #245, opened 2025-07-07
Add beacon_committee_selections endpoint for DVs

KaloyanTanev:
Support endpoint https://ethereum.github.io/beacon-APIs/#/Validator/submitBeaconCommitteeSelections

### DIFF
--- a/multi/beaconcommitteeselections.go
+++ b/multi/beaconcommitteeselections.go
@@ -28,13 +28,13 @@ func (s *Service) BeaconCommitteeSelections(ctx context.Context,
 	*api.Response[[]*apiv1.BeaconCommitteeSelection], error,
 ) {
 	res, err := s.doCall(ctx, func(ctx context.Context, client consensusclient.Service) (any, error) {
-		aggregatedSelections, err := client.(consensusclient.BeaconCommitteeSelectionsProvider).
+		beaconCommitteeSelections, err := client.(consensusclient.BeaconCommitteeSelectionsProvider).
 			BeaconCommitteeSelections(ctx, opts)
 		if err != nil {
 			return nil, err
 		}
 
-		return aggregatedSelections, nil
+		return beaconCommitteeSelections, nil
 	}, nil)
 	if err != nil {
 		return nil, err


service.go
```diff
@@ -366,6 +366,18 @@ type BeaconCommitteeSubscriptionsSubmitter interface {
 	SubmitBeaconCommitteeSubscriptions(ctx context.Context, subscriptions []*apiv1.BeaconCommitteeSubscription) error
 }
 
+// BeaconCommitteeSelectionsProvider is the interface for submitting beacon committee subnet selections requests.
```
Bez625:
```suggestion
// BeaconCommitteeSelectionsProvider is the interface for obtaining beacon committee subnet selections.
```

service.go
```diff
@@ -366,6 +366,18 @@ type BeaconCommitteeSubscriptionsSubmitter interface {
...
+// Used by DV middleware clients.
+type BeaconCommitteeSelectionsProvider interface {
+	// BeaconCommitteeSelections submits beacon committee selections.
```
Bez625:
```suggestion
	// BeaconCommitteeSelections obtains beacon committee selections.
```

mock/beaconcommitteeselections.go
```diff
@@ -0,0 +1,33 @@
...
+package mock
+
+import (
+	"context"
+
+	"github.com/attestantio/go-eth2-client/api"
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+)
+
+// BeaconCommitteeSelections submits beacon committee selections.
```
Bez625:
```suggestion
// BeaconCommitteeSelections obtains beacon committee selections.
```

multi/beaconcommitteeselections.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	consensusclient "github.com/attestantio/go-eth2-client"
+	"github.com/attestantio/go-eth2-client/api"
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+)
+
+// BeaconCommitteeSelections submits beacon committee selections.
```
Bez625:
```suggestion
// BeaconCommitteeSelections obtains beacon committee selections.
```

http/beaconcommitteeselections.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, errors.Join(errors.New("failed to encode beacon committee selections"), err)
+	}
+
+	endpoint := "/eth/v1/validator/beacon_committee_selections"
+	query := ""
+
+	httpResponse, err := s.post(ctx,
+		endpoint,
+		query,
+		&api.CommonOpts{},
```
Bez625: Instead of creating a new struct I think this should probably come from opts.Common?

http/beaconcommitteeselections.go
```diff
@@ -0,0 +1,66 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/attestantio/go-eth2-client/api"
+	apiv1 "github.com/attestantio/go-eth2-client/api/v1"
+)
+
+// BeaconCommitteeSelections submits beacon committee selections.
```
Bez625:
```suggestion
// BeaconCommitteeSelections obtains beacon committee selections.
```

http/beaconcommitteeselections.go
```diff
@@ -0,0 +1,66 @@
...
+	opts *api.BeaconCommitteeSelectionsOpts,
+) (
+	*api.Response[[]*apiv1.BeaconCommitteeSelection],
+	error,
+) {
+	if err := s.assertIsSynced(ctx); err != nil {
+		return nil, err
+	}
+
+	specJSON, err := json.Marshal(opts.Selections)
```
Bez625, line 42: We should validate opts is not nil first since it's required here, e.g.

```golang
  if opts == nil {
      return nil, errors.New("no beacon committee selections options specified")
  }
```

multi/beaconcommitteeselections.go
```diff
@@ -0,0 +1,49 @@
...
+func (s *Service) BeaconCommitteeSelections(ctx context.Context,
+	opts *api.BeaconCommitteeSelectionsOpts,
+) (
+	*api.Response[[]*apiv1.BeaconCommitteeSelection], error,
+) {
+	res, err := s.doCall(ctx, func(ctx context.Context, client consensusclient.Service) (any, error) {
+		aggregatedSelections, err := client.(consensusclient.BeaconCommitteeSelectionsProvider).
```
Bez625: Why the variable name `aggregatedSelections`? doCall `carries out a call on the active clients in turn until one succeeds.` so we should only get responses from one client. This feels misleading.

http/beaconcommitteeselections.go
KaloyanTanev: I've set `client.ErrNoOptions` instead of `errors.New("no beacon committee selections options specified")`, as I saw it being used in such scenarios for other calls.

multi/beaconcommitteeselections.go
KaloyanTanev: I chose the `aggregatedSelections` name as what a validator client would send to the distributed validator client is a "partial" selection. Then the DV will "aggregate" all the partial selections and return to the VC the "aggregated" selections, hence the naming. I can go with something different if it's misleading. My perspective is a bit biased towards DVs, so it may be indeed misleading.
`fullSelections`?
`beaconCommitteeSelections`?

Bez625: I think beaconCommitteeSelections is good

http/beaconcommitteeselections.go
Bez625: Great!

PR #256, opened 2025-08-08
Use dedicated fulu package types instead of electra types

hoanhan101:


KaloyanTanev:
If the block structure is unchanged, is this needed?

PR #270, opened 2025-09-23
Switch to dynssz codegen

### LINKED ISSUE #259: Dependency on outdated unstable release of dynamic-ssz
This project depends on dynamic-ssz v0.0.4:

https://github.com/attestantio/go-eth2-client/blob/e8c9fbe66ad30e880b4635d80d78a694e90d8967/go.mod#L14

That project already released an stable v1. Please update this dependency to the most recent version, currently v1.0.1.

I am [packaging mev-boost in Debian](https://collective.flashbots.net/t/mev-boost/5060#p-10282-debian-packaging-1). I'm interested in this new version because it replaces an [unmaintained dependency](https://github.com/pk910/dynamic-ssz/issues/11). According to the changelog, the new versions also come with performance improvements.

pk910:
This PR changes the code generator for ssz code to dynamic-ssz.

The code generated by dynamic-ssz has equal performance characteristics as the code generated by fastssz.

Biggest benefit of switching to the dynssz codegen:  no more pre- and post work to get the damn ssz code generated. The code generator just works out of the box 🎉. It uses types properly and does not require manual edits, which is a huge pain with the existing fastssz code generator.

supersedes #260, #242
fixes #259


api/v1/blobs_ssz.go
```diff
@@ -1,99 +1,72 @@
...
-		if err != nil {
-			return err
-		}
-		*b = make(Blobs, num)
-		for ii := 0; ii < num; ii++ {
-			if (*b)[ii] == nil {
-				(*b)[ii] = new(deneb.Blob)
-			}
-			copy((*b)[ii][:], buf[ii*131072:(ii+1)*131072])
+func (t Blobs) UnmarshalSSZ(buf []byte) (err error) {
```
Bez625: Is there a reason this has gone from a pointer receiver to a value receiver? I suspect this won't deserialise properly as we're

api/v1/blobs_ssz.go
```diff
@@ -1,99 +1,72 @@
...
+		if t[i] == nil {
+			t[i] = new(deneb.Blob)
 		}
+		buf := buf[131072*i : 131072*(i+1)]
+		copy(t[i][:], buf)
 	}
 	return nil
 }
 
-// SizeSSZ returns the ssz encoded size in bytes for the Blobs object
```
Bez625, line 54: Am I correct in thinking that dynamic-ssz does not include doc-strings? If not is there a plan to add them?

Bez625 (changes requested):

api/v1/blobs_ssz.go
pk910: Yea, this is a bug that I've discovered recently too.
It affects lists only, structs are already passed as pointer receiver.
rel. https://github.com/pk910/dynamic-ssz/pull/147 (fixed as part of that PR)

api/v1/blobs_ssz.go
pk910: Yea, I've recently added some default doc strings:
https://github.com/pk910/dynamic-ssz/pull/145

pk910:
@Bez625 
I've bumped the dynamic-ssz lib to release 1.3.0 and re-generated the ssz code.
This fixes both review comments :)

pk910:
@Bez625 
I've bumped the dynamic-ssz lib to release 1.3.1 and re-generated the ssz code (just semantic changes).

The 1.3.1 release mainly contains a new config format for the generator that simplifies the utterly long cli args in the generate.go
There's now a separate `generate.yaml` that lives along each `generate.go` and contains the types list in a more readable format.

PR #282, opened 2026-03-15
Add json and yaml serialization for syncaggregatorselectiondata

jshufro:


spec/altair/syncaggregatorselectiondata_test.go
```diff
@@ -0,0 +1,121 @@
+// Copyright © 2021 Attestant Limited.
```
AntiD2ta, line 1:
```suggestion
// Copyright © 2026 Attestant Limited.
```

spec/altair/syncaggregatorselectiondata_test.go
```diff
@@ -0,0 +1,121 @@
...
+	tests := []struct {
+		name  string
+		input []byte
+		err   string
+	}{
+		{
+			name:  "Good",
+			input: []byte(`{slot: 1, subcommittee_index: 3}`),
+		},
+	}
```
AntiD2ta, line 103: The JSON tests look great, covering many cases. Could we please add more cases here, even if JSON and YAML methods share `unpack`? Some cases that might be worth covering are:

- Empty/invalid YAML input
- Missing fields (YAML unmarshaler may behave differently than JSON for zero-values vs absent keys)

spec/altair/syncaggregatorselectiondata_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"testing"
+
+	"github.com/attestantio/go-eth2-client/spec/altair"
+	"github.com/goccy/go-yaml"
+	"github.com/stretchr/testify/assert"
+	require "github.com/stretchr/testify/require"
```
AntiD2ta, line 24: This is rather a nitpicky comment. Since other test files already use the plain import, can we aim for consistency here?

```suggestion
	"github.com/stretchr/testify/require"
```

AntiD2ta (changes requested):
Thanks a lot for the PR @jshufro !

I requested changes mainly for the test cases comment. The other two comments should be covered by a linter we will introduce soon.

Please also fix the linter reporting issues (we raised and merged a PR related to the linter issues that were not introduced in this PR. If you keep seeing pre-existing issues, try rebasing your branch with master).

AntiD2ta:
We will ignore the failing `http-tests` CI because it only runs correctly if the PR doesn't come from a fork. This PR doesn't affect the http package anyway.

PR #286, opened 2026-03-25
Add provider address to event data structs

AntiD2ta:
## Changes

- Added `Provider string` field to all owned event data structs: `HeadEvent`, `BlockEvent`, `ChainReorgEvent`, `FinalizedCheckpointEvent`, `BlobSidecarEvent`, `BlockGossipEvent`, `PayloadAttributesEvent`, `DataColumnSidecarEvent`, and the generic `Event` container
- Added `WithProvider`/`Provider` context helpers in `api/v1` for universal provider extraction across all event types
- HTTP and multi-client event handlers now populate the provider address on event data and context before forwarding to consumers

Note that seven event types use `spec` structs that model consensus data types (`VersionedAttestation`, `AttesterSlashing`, etc.) that should not carry a `Provider` field. 

The context-based approach (`apiv1.WithProvider`/`apiv1.Provider`) provides a universal mechanism that works for all 15 event types, while the struct field is a convenience for the 9 types we own.

## Example usage on Vouch

When the provider is available directly on the HeadEvent struct:

```go
  func (s *Service) HandleHeadEvent(ctx context.Context, event *apiv1.HeadEvent) {
      // The provider address is embedded in the event data.
      provider := event.Provider // e.g. "http://localhost:5052"

      s.blockReceiptDelay.WithLabelValues(provider).Observe(delay.Seconds())
  }
```

  For event types that use external spec structs (e.g. attestations), where the struct can't carry the field, we can extract it from context instead:

```go
  func (s *Service) HandleAttestationEvent(ctx context.Context, event *spec.VersionedAttestation) {
      provider := apiv1.Provider(ctx) // universal mechanism, works for all event types

      // ...
  }
```

  The context mechanism (apiv1.Provider(ctx)) works for all 15 event types, so Vouch could use it universally if preferred. The struct field is a convenience for the 8 owned types.


api/v1/provider.go
```diff
@@ -0,0 +1,37 @@
...
+
+import "context"
+
+type providerKey struct{}
+
+// WithProvider returns a context with the provider address set.
+// This is the universal mechanism for obtaining the provider address
+// across all event types.  Event structs owned by this package also
+// carry the address directly in their Provider field.
+func WithProvider(ctx context.Context, provider string) context.Context {
```
Bez625: I'm not a fan of the provider being passed in the context. I looked at vouch and I can only see us using head and block event handlers. I think let's limit this to updating the `*Event` structs to have the `Provider` field and set them where possible. Given we have explicit types at the callsite for these, we know if they have the `Provider` or not so it's type safe / requires no validation. 

I think if we want this behaviour on the other handlers we should look to find reasons and upgrade them to returning `*Event` types that explicitly set the Provider.

Bez625 (changes requested):

api/v1/provider.go
AntiD2ta: Done [346dc29](https://github.com/attestantio/go-eth2-client/pull/286/commits/346dc296d0150ec114c7537f34133a320b3aedf6)

AntiD2ta:
Created https://github.com/attestantio/go-eth2-client/issues/288 to address adding the Provider on consensus types event data structs.

api/v1/provider.go
AntiD2ta: Also: https://github.com/attestantio/go-eth2-client/pull/286#issuecomment-4136184219

PR #291, opened 2026-04-15
Switch fastssz to dynamic-ssz

Bez625:
This PR changes the code generator for ssz code to dynamic-ssz.

The code generated by dynamic-ssz has equal performance characteristics as the code generated by fastssz.


api/v1/blobs_ssz.go
```diff
@@ -1,99 +1,91 @@
...
-		hh.MerkleizeWithMixin(subIndx, numItems, 128)
 	}
-
-	return
-}
-
-// GetTree ssz hashes the Blobs object
-func (b *Blobs) GetTree() (*ssz.Node, error) {
-	return ssz.ProofTree(b)
+	hh.Merkleize(idx)
```
AntiD2ta: **Critical — `Blobs.HashTreeRoot` lost its `List[Blob, 72]` length mix-in.**

Master computed this as `hh.MerkleizeWithMixin(subIndx, numItems, 128)` (List semantics — the element count is folded into the root). Here it is a plain `hh.Merkleize(idx)` with no mix-in, i.e. Vector semantics, so `Blobs.HashTreeRoot()` now returns a **different root** for identical blob data. Any cache key, equality check against a beacon node, or proof verification breaks after upgrade.

**Root cause:** `type Blobs []*deneb.Blob` is a bare named slice — there is no struct field to hang an `ssz-max` tag on, so the `72` bound was never expressed in the type. In the fastssz era it survived only because `blobs_ssz.go` was hand-edited and excluded from codegen; this PR regenerates the file from the untagged type, so dynssz has no tag to read.

**Fix:** follow the pattern this repo already uses for `BlobSidecars` in `api/blobsidecars.go` — drop `Blobs` from the `dynssz-gen -types` / `rm -f` lists and hand-write the methods against a tagged wrapper:

```go
type blobsSSZ = dynssz.TypeWrapper[struct {
    Blobs []*deneb.Blob `ssz-max:"72"`
}, []*deneb.Blob]
```

dynssz then restores both the length mix-in (here) and the max-72 cap (`UnmarshalSSZ`, separate comment), with no change to the public `type Blobs []*deneb.Blob`.

api/v1/blobs_ssz.go
```diff
@@ -1,99 +1,91 @@
...
-		*b = make(Blobs, num)
-		for ii := 0; ii < num; ii++ {
-			if (*b)[ii] == nil {
-				(*b)[ii] = new(deneb.Blob)
-			}
-			copy((*b)[ii][:], buf[ii*131072:(ii+1)*131072])
+// UnmarshalSSZ unmarshals the *Blobs from SSZ-encoded bytes.
+func (t *Blobs) UnmarshalSSZ(buf []byte) (err error) {
+	itemCount := len(buf) / 131072
+	if len(buf)%131072 != 0 {
```
AntiD2ta: **Critical — `Blobs.UnmarshalSSZ` dropped the max-72 cap.**

Master used `ssz.DivideInt2(len(buf), 131072, 72)`, rejecting any buffer representing more than 72 blobs with `ErrListTooBig`. This version only checks byte-alignment (`len(buf)%131072`) and accepts any item count, so a malformed/malicious `/eth/v1/beacon/blobs` response with more than 72 entries is silently decoded as valid instead of rejected.

Same root cause and fix as the `HashTreeRoot` mix-in loss (separate comment on line 89): expressing `ssz-max:"72"` via a `dynssz.TypeWrapper` (mirroring `api/blobsidecars.go`) restores the cap here too.

spec/deneb/kzgcommitmentinclusionproof.go
```diff
@@ -24,7 +24,7 @@ import (
 const kzgCommitmentProofElements = 17
 
 // KZGCommitmentInclusionProof is the proof of inclusion for a KZG commitment.
-type KZGCommitmentInclusionProof [kzgCommitmentProofElements]KZGCommitmentInclusionProofElement
+type KZGCommitmentInclusionProof []KZGCommitmentInclusionProofElement
```
AntiD2ta, line 28: **Critical — an unset `KZGCommitmentInclusionProof` now marshals to JSON `null`.**

Changing `[17]…Element` → `[]…Element` makes a nil proof serialize as `null` (stdlib slice behavior; there is no custom `MarshalJSON`, and `BlobSidecar.MarshalJSON` copies the field verbatim). A `BlobSidecar` built programmatically without populating the proof now emits `"kzg_commitment_inclusion_proof":null` instead of a 17-element array; re-decoding that output fails with `"incorrect number of elements"`. The old fixed array always serialized as 17 hex strings.

`UnmarshalJSON` was patched with `make(…, 17)` so wire-decoded values are fine — the gap is on the marshal side. Consider a custom `MarshalJSON` that rejects/zero-fills a non-17-length proof, or keeping the fixed-size array.

api/blobsidecars.go
```diff
@@ -15,33 +15,50 @@ package api
...
+	Sidecars []*deneb.BlobSidecar `ssz-max:"72"`
+}, []*deneb.BlobSidecar]
+
 // UnmarshalSSZ ssz unmarshals the BlobSidecars object.
-// This is a hand-crafted function, as automatic generation does not support immediate arrays.
 func (b *BlobSidecars) UnmarshalSSZ(buf []byte) error {
-	num, err := ssz.DivideInt2(len(buf), 131928, 72)
-	if err != nil {
+	blobs := blobSidecarsSSZ{}
+	if err := dynssz.GetGlobalDynSsz().UnmarshalSSZ(&blobs, buf); err != nil {
```
AntiD2ta: **Critical — `BlobSidecars` SSZ routes through the mainnet-only global dynssz.**

All four SSZ methods (this one plus `MarshalSSZ`, `SizeSSZ`, `HashTreeRoot`) use `dynssz.GetGlobalDynSsz()`, a singleton only ever built with nil (mainnet) specs and never configured via `SetGlobalSpecs`. The HTTP service otherwise decodes blocks/states with a per-request `dynssz.NewDynSsz(specs.Data)`.

A client built with `http.WithCustomSpecSupport(true)` against a non-mainnet preset therefore decodes sidecars against mainnet dimensions — the exact scenario custom-spec support was requested for. Consider threading the spec-aware instance into `BlobSidecars` SSZ instead of the global.

spec/altair/synccommitteecontribution.go
```diff
@@ -33,7 +33,7 @@ type SyncCommitteeContribution struct {
 	BeaconBlockRoot   phase0.Root `ssz-size:"32"`
 	SubcommitteeIndex uint64
 	// AggregationBits size is SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT
-	AggregationBits bitfield.Bitvector128 `dynssz-size:"SYNC_COMMITTEE_SIZE/SYNC_COMMITTEE_SUBNET_COUNT" ssz-size:"16"`
+	AggregationBits bitfield.Bitvector128 `dynssz-size:"SYNC_COMMITTEE_SIZE/4/8" ssz-size:"16"`
```
AntiD2ta, line 36: **Medium — `SYNC_COMMITTEE_SUBNET_COUNT` hardcoded to the literal `4`.**

```diff
- dynssz-size:"SYNC_COMMITTEE_SIZE/SYNC_COMMITTEE_SUBNET_COUNT"
+ dynssz-size:"SYNC_COMMITTEE_SIZE/4/8"
```

The point of `dynssz-size` is to stay correct under custom presets, but baking in `4` defeats that: under a preset where `SYNC_COMMITTEE_SUBNET_COUNT != 4` the bitvector length is computed wrong (encode/decode/HTR mismatch). Correct for mainnet/minimal (where it is 4). This looks forced by `-without-dynamic-expressions` — worth confirming whether the expression syntax can keep `…/SYNC_COMMITTEE_SUBNET_COUNT/8` instead of a literal.

api/blobsidecars.go
```diff
@@ -15,33 +15,50 @@ package api
...
+// MarshalSSZ ssz marshals the BlobSidecars object.
+func (b *BlobSidecars) MarshalSSZ() ([]byte, error) {
+	return dynssz.GetGlobalDynSsz().MarshalSSZ(&blobSidecarsSSZ{
+		Data: b.Sidecars,
+	})
+}
+
+// SizeSSZ returns the size of the BlobSidecars object.
+func (b *BlobSidecars) SizeSSZ() int {
+	size, _ := dynssz.GetGlobalDynSsz().SizeSSZ(&blobSidecarsSSZ{
```
AntiD2ta, line 60: **Medium — `SizeSSZ` swallows its error and returns 0.**

`size, _ := dynssz.GetGlobalDynSsz().SizeSSZ(…)` silently returns `0` if sizing fails. A caller using this to pre-size a buffer or validate an expected length gets `0` with no signal. The method signature can't return an error, so at minimum a comment (or a panic on the "impossible" error) would be clearer than a silent discard.

AntiD2ta (approved):

PR #298, opened 2026-07-14
dynssz bump & UX improvements

pk910:


http/blobs.go
```diff
@@ -86,7 +86,8 @@ func (s *Service) blobsFromSSZ(ctx context.Context, res *httpResponse) (*api.Res
 		if specs, err = s.Spec(ctx, &api.SpecOpts{}); err != nil {
 			return nil, errors.Join(errors.New("failed to request specs"), err)
 		}
-		err = response.Data.UnmarshalSSZDyn(dynssz.NewDynSsz(specs.Data), res.body)
+		dynSsz := dynssz.NewDynSsz(specs.Data)
+		err = dynSsz.UnmarshalSSZ(response.Data, res.body)
```
AntiD2ta:
```suggestion
		err = dynSsz.UnmarshalSSZ(&response.Data, res.body)
```

`dynSsz.UnmarshalSSZ` is a generic function that requires a pointer target to write the decoded value back into. Here it's handed `response.Data`, which is an `apiv1.Blobs (i.e. []*deneb.Blob)` slice value and on this path it is nil, since it's only initialised on the empty-body early return.

I'm concerned that any client built with `WithCustomSpecSupport(true)` that calls `Blobs()` against a node returning SSZ would get "failed to decode blobs" and zero blobs. 

The mainnet path on line 92 (http/blobsidecars.go) works because a pointer-receiver method auto-takes the address of the addressable field, routing through the generic function loses that. The sibling http/blobsidecars.go does it correctly: `data := &api.BlobSidecars{}`, then `dynSsz.UnmarshalSSZ(data, …).`

api/blobsidecars.go
```diff
@@ -15,58 +15,12 @@ package api
...
-	return size
-}
-
-// HashTreeRoot ssz hashes the BlobSidecars object.
-func (b *BlobSidecars) HashTreeRoot() ([32]byte, error) {
-	return dynssz.GetGlobalDynSsz().HashTreeRoot(&blobSidecarsSSZ{
-		Data: b.Sidecars,
-	})
-}
+var _ = sszutils.Annotate[BlobSidecars](`ssz-type:"wrapper"`)
```
AntiD2ta, line 26: The wrapper annotation carries `ssz-type:"wrapper"` but no `ssz-max:"72"`, and the Sidecars field is untagged. 

On our branch, we spotted the following consequences of this:

- **Wrong root:** `HashTreeRootWith` calls plain `hh.Merkleize(idx)` (blobsidecars_ssz.go:105) with no limit and, crucially, no length mix-in. It no longer yields the consensus `List[BlobSidecar, 72]` root, so it disagrees with every other client on the network.
- **No decode cap:** `UnmarshalSSZ` only checks byte-alignment (`ErrListNotAlignedFn(len(buf), 131928))`; a body carrying >72 sidecars is silently accepted instead of returning `ErrListTooBig`. (Note `var _ = sszutils.ErrListTooBig` is declared but unused for this field.)

Do you spot any problem in carrying `ssz-max:"72"` through the wrapper annotation so the generator emits the limit?

api/v1/blobs_ssz.go
```diff
@@ -1,78 +1,94 @@
...
 
-// UnmarshalSSZDyn ssz unmarshals the Blobs object using the supplied dynamic SSZ
-// instance, allowing the caller to decode against a custom (non-mainnet) spec.
-func (b *Blobs) UnmarshalSSZDyn(dynSSZ *dynssz.DynSsz, buf []byte) error {
-	wrapper := blobsSSZ{}
-	if err := dynSSZ.UnmarshalSSZ(&wrapper, buf); err != nil {
-		return err
+// UnmarshalSSZ unmarshals the *Blobs from SSZ-encoded bytes.
+func (t *Blobs) UnmarshalSSZ(buf []byte) (err error) {
+	itemCount := len(buf) / 131072
```
AntiD2ta, line 43: Similar to https://github.com/attestantio/go-eth2-client/pull/298/changes#r3588776026, on `Blobs: UnmarshalSSZ` validates only alignment (no > 72 cap), and `HashTreeRootWith` ends in `hh.Merkleize(idx)`(blobs_ssz.go:92) with no length mix-in, so the hash-tree root is Vector-shaped, not `List[Blob, 72]`-shaped.

The root cause is structural: `type Blobs []*deneb.Blob` (blobs.go:23) is a bare named slice with no struct field and no `Annotate` marker, so there is nowhere for `ssz-max:"72"` to attach. We solved this problem in PR #291 with a `TypeWrapper`.

Wdyt of wrapping `Blobs` the same way `BlobSidecars` is wrapped and attach `ssz-max:"72"` to that wrapper?

api/v1/generate.go
AntiD2ta, line 1: Is `blobs_ssz.go` missing on line 16?

`generate.yaml` now emits `Blobs` → `blobs_ssz.go`, but that file isn't in the pre-clean `rm -f` list. The only such mismatch across all 18 regenerated packages, and asymmetric with api/generate.go:17, which did add its parallel new output `blobsidecars_ssz.go`.

This is probably harmless today (the generator overwrites in place), it might bite later if `Blobs` is renamed or removed, thus `blobs_ssz.go` is then left orphaned but still compiling, diverging from the type set.

AntiD2ta (changes requested):

api/blobsidecars.go
pk910: oh, yea, that's right.  I missed the `ssz-max` annotation, which should go to the `BlobSidecars.Sidecars` field now.
The annotation on the outer `BlobSidecars` container tells the lib to serialize the inner field only, skipping the container offset prefix.

api/v1/blobs_ssz.go
pk910: Yea,  similar to the issue above,  I missed adding the ssz-max annotation, sorry for that.

The TypeWrapper approach you had before is working perfectly fine, and is exactly the intended implementation way for such cases in the versions before 1.3.0.
However, it feels a bit "dirty", as we're technically creating dummy wrappers and copying around the data field in memory (even if it's just a pointer), have the manual ssz handlers, etc.

So, with 1.3.0 I've added `sszutils.Annotate`, which allows annotations for any type without the need for the struct wrapper.
That basically removes the need for these TypeWrapper containers, as the type can directly be annotated.

So yea,  nothing wrong with your approach,  just feels a bit cleaner with the new annotation way imo
(happy to revert if you prefer the other way - the TypeWrapper won't be removed)

http/blobs.go
pk910: good catch :)

One side note:
The current way of creating a new DynSSZ instance for each serialization operation is not very good.
There are type caches living inside the instance that are quite expensive to create (reflection based type traversal), but static per type & specs object.
This is currently not a big issue, as the dynamic serialization is only used with `WithCustomSpecSupport(true)`,  but might cause some unnecessary extra load.
To make this nice,  the dynssz instance should be stored in the `Service` struct and reused for all requests. Only re-initialize when the spec is re-fetched

AntiD2ta (approved):

PR #302, opened 2026-07-23
Gloas support for Glamsterdam devnet-7 (types + transport)

AntiD2ta:
## Glamsterdam / Gloas support

Additively ports **Gloas** (Glamsterdam devnet-7) support from the
[ethpandaops fork](https://github.com/ethpandaops/go-eth2-client) into
attestantio, imports rewritten `ethpandaops` → `attestantio`, additive only.

Target spec: consensus `v1.7.0-alpha.12`. **Base branch: `gloas` (never master).**

Two slices accumulate on this branch and ship as one PR: the **types layer** and
the **transport layer**. Both are now in.

---

## Slice 1 — types layer

Additively ports the Gloas type surface.

### What landed
- **`spec/version` restructure**: `DataVersion`/`BuilderVersion` moved to a leaf
  `spec/version` package with an alias shim in package `spec` (`spec/version.go`).
  Added `DataVersionGloas`. Heze omitted (out of scope).
- **`spec/gloas`** package (92 source files) — types + JSON/YAML/SSZ codecs +
  ported roundtrip tests. SSZ regenerated with `dynssz-gen`, byte-identical to the
  fork output.
- **Gloas arms** added to 10 existing `Versioned*` wrappers (additive only).
- **4 new ePBS versioned wrappers**: `VersionedExecutionPayloadEnvelope`,
  `VersionedSignedExecutionPayloadEnvelope`, `VersionedSignedExecutionPayloadBid`,
  `VersionedExecutionRequests`.
- **`api/v1/gloas`**: `SignedExecutionPayloadEnvelopeContents`.

### Key decisions
- `spec/version.go` uses **const** aliases (preserving prior const semantics) and
  **preserves `spec.DataVersionFromString`** (the fork drops it) — additive.
- **Kept the fork's `ssz-index` struct tags**: they are load-bearing for
  `dynssz-gen` (removing them changes the generated SSZ — verified empirically).
- **Excluded fork breaking changes**: no `BaseFeePerGas` →
  `BaseFeePerGasLE` rename; `VersionedSignedBeaconBlock.ExecutionRequests()` keeps
  its `*electra.ExecutionRequests` return (gloas arm returns an explicit error).
- Lint: `//nolint:gocyclo` on version-switch methods that crossed the complexity
  threshold once the gloas case was added (matches existing repo convention); a
  scoped `.golangci.yml` line-length exclusion for the gloas type packages (their
  ePBS `dynssz` size-expression tags exceed 132 chars once gofmt-aligned and
  cannot be wrapped).

---

## Slice 2 — transport layer (ePBS endpoints, events, client parity)

Additively ports the ePBS transport surface.

### What landed
- **3 ePBS HTTP endpoints** (`http/`): `SignedExecutionPayloadEnvelope` (GET),
  `SubmitExecutionPayloadBid` (POST), `SubmitExecutionPayloadEnvelope` (POST,
  stateless Contents form), with `api/` opts types and 3 provider interfaces
  (`ExecutionPayloadProvider`, `ExecutionPayloadEnvelopeSubmitter`,
  `ExecutionPayloadBidSubmitter`).
- **7 new SSE event topics** with typed handlers (`execution_payload`,
  `execution_payload_available`, `execution_payload_bid`,
  `execution_payload_gossip`, `fast_confirmation`, `payload_attestation_message`,
  `proposer_preferences`); 3 new `api/v1` event payload types
  (`ExecutionPayloadEvent`, `ExecutionPayloadAvailableEvent`,
  `FastConfirmationEvent`) + an `unmarshalVersionedEventData` helper that accepts
  both `{version,data}`-wrapped and bare payloads.
- **multi / mock / testclients** parity for the 3 endpoints.
- **New HTTP tests** for the 3 endpoints (success + error paths), 3 `api/v1`
  event round-trip tests, and `TestEventHandler` extended with the 7 new topics
  (wrapped + bare payloads).
- Shared http test service now built with `WithCustomSpecSupport(true)` (the
  interim devnet runs the minimal preset).

### Key decisions
- Dropped the fork's `spec/all` `Agnostic*` endpoint/opts/interface variants
  (spec/all not adopted) and all Heze arms (out of scope); the non-agnostic
  variants use the versioned wrappers from slice 1.
- Added a local `dynSSZForRequest` helper mirroring the repo's existing inline
  dynssz pattern (`GetGlobalDynSsz` for mainnet, `NewDynSsz(spec)` for custom
  presets) rather than the fork's Service-cached version, which depends on
  fields this tree lacks.
- **Two wire details intentionally diverge from the (now-stale) fork to match the
  live beacon-APIs spec** (verified against upstream `master`): the
  submit-envelope header is `Eth-Blob-Data-Included: true` (the fork used the
  removed `Eth-Execution-Payload-Blinded: false`), and `FastConfirmationEvent`
  carries `current_slot` (the fork omitted it; parsed tolerantly).


AntiD2ta:
## `http-tests-gloas` is expected to fail — full accounting

> **Amended.** Two claims in the first version of this comment were wrong. They are corrected in place below rather than quietly removed, so that anyone who read the original can see what changed:
>
> 1. **The failing set is client-dependent, so the list below describes one run rather than the general case.** The endpoint is a load balancer over ~47 nodes spanning six consensus implementations, and the caller cannot pin one. The same commit against the same address gave **15** failing subtests on two runs and a different **20** on a third, with the chain healthy throughout.
> 2. **The timeout was not caused by runner bandwidth.** That reasoning is disproved below.
>
> The suite now prints `endpoint identity: status=… version=… upstream=…` before any test runs, so each run records which implementation answered it.

The advisory second job on this branch runs the whole `./http/...` suite against a Gloas devnet with `HTTP_REQUIRE_GLOAS=true`, which turns every fork gate from a skip into a failure. It is `continue-on-error: true` by design, so it does not block this PR — but it is not noise either, so here is every failure and why.

The endpoint is a public Glamsterdam devnet-7 beacon API: mainnet preset, `GLOAS_FORK_EPOCH: 38`, head currently around epoch 3870, unauthenticated. **On a Prysm upstream** 15 subtests fail, in four groups. **Only one is a defect in this library.**

### 1. A real bug here — 1 failure

`TestBeaconState/Head_(json)`

```
builders: invalid JSON: json: cannot unmarshal string into Go struct field builderJSON.version of type uint8
```

`gloas.Builder.version` is serialised three mutually incompatible ways. beacon-APIs `types/primitive.yaml` declares `Uint8` as a decimal string (`pattern: ^[1-2]?[0-9]{1,2}$`, example `"0"`); Prysm emits a 0x-prefixed hex string (`"version":"0x00"`); this library emits and expects a bare JSON number. `encoding/json` will not coerce a string into a numeric field, so a single field makes the *entire* Gloas beacon state undecodable over JSON. Worth noting `json:"version,string"` would not fix it — that accepts decimal strings only, not `0x00`.

### 2. Optional header absent from an SSZ response — 3 failures

`TestEPBSProposal/SSZPayloadExcluded`, `TestEPBSProposal/SSZPayloadIncluded`, `TestSubmitProposalGloas`

```
no Eth-Execution-Payload-Included header in epbs proposal response
```

`apis/validator/block.v4.yaml` declares `Eth-Execution-Payload-Included` **optional**, and also exposes `execution_payload_included` as a response body field. For JSON that is harmless, and on a Prysm upstream the JSON subtests pass. For SSZ the header is the only possible in-band signal, and Prysm sends neither it nor `Eth-Execution-Payload-Value` — checked under both `Accept: application/octet-stream` and `Accept: application/json`, in case it were set only for SSZ.

The client refuses to guess on purpose: the flag selects which container the body holds, so defaulting would decode block contents as a bare block and discard the envelope, blobs and proofs the caller needs to publish. Being raised upstream.

### 3. Node-side 500s — 6 failures

`TestAttestationPool` (3 subtests) and `TestAttestationPoolCommitteeIndexSet`:

```
500 Unable to convert attestation of type *eth.AttestationGloas
```

`TestBlobsSidecars/Head_Block_Root` and `/Head_Block_Slot`:

```
500 blobs from data columns: ... data column sidecar is not a fulu type
```

Both are server-side. The first also masks a gap on our side: `verifyAttestationPool` has no Gloas arm and would fail even against a fixed server — the 500 simply arrives first.

### 4. Expected on a Gloas network — 5 failures

`TestProposal/Good`, `TestSubmitBeaconBlock/Good`, `TestSubmitProposal/Good` — post-Gloas block production is v4-only; v3 returns an empty 200, so these fail with `unexpected end of JSON input`. Permanent on a Gloas node until the v3 path reports this explicitly rather than failing to parse.

`TestValidatorBalances/SingleGenesisIndex` — expects 32 ETH, node reports 1024 ETH. This devnet stakes validators at 1024 ETH; the assertion encodes a network assumption.

`TestAggregateAttestation/NotFound` — expects 404, gets 500. Previously observed as non-deterministic on other endpoints too.

### Why the set grows on a non-Prysm upstream

A third run of the same commit produced 20 failures instead of 15, and the newcomers are what a different implementation produces: `TestEPBSProposal/JSONPayloadExcluded` and `/JSONPayloadIncluded`, which had passed in every prior run; `TestExecutionPayloadEnvelope/PreGloasSlotIsNotSentinel`; `TestPayloadAttestationData/PastSlot`; `TestSubmitPayloadAttestationMessages/ReachesServer`; and an error-message-format group — `TestProposerDuties/Epoch`, `TestValidatorLiveness/Good`, `TestSubmitAttestations/Good`, `TestAttestationData/BadSlot`.

Degradation is ruled out: head was at epoch 3874 with finalized 3872, 58 peers, not syncing, not optimistic. No query parameter or header selects an upstream, and there is no per-client hostname, so this cannot currently be pinned from the client side. Read a change in the failing set against the `endpoint identity` line before treating it as a regression.

### On the job's timeout

An earlier run ended in `panic: test timed out after 10m0s` with `TestValidators` at 4m12s. The job's timeout is now 25m, and the reason is **variance in which upstream the load balancer assigns**, not bandwidth: the same commit against the same address spent 1m34s on one `TestValidators` subtest and blew a 10m budget, then did the whole test in 11.37s with the suite finishing in 103s, while the chain advanced continuously and only the upstream identity changed.

Runner bandwidth is specifically not the constraint — the other endpoint in this workflow serves a validator set 2.7x larger, four times over, inside a 2m49s job.

Separately, three `TestValidators` rows no longer fetch every validator to prove that a state id resolves, which removes ~69s of transfer per run on a mainnet-scale network. One row still fetches the whole set deliberately, as the only place in the suite that decodes a validator list at that scale.

spec/gloas/indexedattestation.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Sprintf("ERR: %v", err)
+	}
+
+	return string(data)
+}
+
+func (i *IndexedAttestation) unpack(indexedAttestationJSON *indexedAttestationJSON) error {
+	var err error
+	// Spec tests contain indexed attestations with empty attesting indices.
+	// if indexedAttestationJSON.AttestingIndices == nil {
```
Bez625: Why is this commented code here?

spec/gloas/attestation.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/goccy/go-yaml"
+	"github.com/pkg/errors"
+)
+
+// Attestation is the Ethereum 2 attestation structure.
+type Attestation struct {
+	AggregationBits bitfield.Bitlist        `ssz-index:"0" ssz-type:"progressive-bitlist"`
+	Data            *phase0.AttestationData `ssz-index:"1"`
+	Signature       phase0.BLSSignature     `ssz-index:"2" ssz-size:"96"`
```
Bez625, line 34: I'm not sure why we need to change Attestation - it seems the only diff between spec.electra.Attestation is the additional metadata around `ssz-index` and `ssz-type`.

Can we not apply this to electra?

spec/gloas/attestation.go
```diff
@@ -0,0 +1,212 @@
...
+	bitfield "github.com/OffchainLabs/go-bitfield"
+	"github.com/attestantio/go-eth2-client/spec/electra"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
+	"github.com/goccy/go-yaml"
+	"github.com/pkg/errors"
+)
+
+// Attestation is the Ethereum 2 attestation structure.
+type Attestation struct {
+	AggregationBits bitfield.Bitlist        `ssz-index:"0" ssz-type:"progressive-bitlist"`
```
Bez625, line 32: The ssz-max bound seems to have been removed as part of this. Is that a change to the spec or the bound has been removed?

spec/gloas/indexedattestation.go
AntiD2ta: Good catch. This was disabled because the spec vectors include empty attesting-index lists, which are valid for the Gloas progressive list. The commented code no longer explains a useful decision, so I have removed it.

spec/gloas/attestation.go
AntiD2ta: Yes, this needs to be a separate Gloas type. The Go fields and JSON happen to look the same, but the SSZ schema does not: Gloas makes Attestation a ProgressiveContainer and aggregation_bits a ProgressiveBitlist. That gives it a different hash-tree root, which feeds the aggregate signing root. Changing Electra would break Electra roots; reusing Electra would sign the wrong Gloas root. The ssz-index and ssz-type tags are therefore load-bearing, not just metadata. The spec calls out both changes: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.12/specs/gloas/beacon-chain.md#attestation

spec/gloas/attestation.go
AntiD2ta: Yes. This is an intentional spec change, not an omitted check. Gloas defines AggregationBits as ProgressiveBitlist rather than the Electra Bitlist[MAX_VALIDATORS_PER_COMMITTEE * MAX_COMMITTEES_PER_SLOT], so the old 131072 bound is no longer part of the SSZ type. The generated Gloas codec therefore must not retain the Electra ssz-max check. See the Gloas type definition and modified Attestation container: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.12/specs/gloas/beacon-chain.md#types

AntiD2ta:
### Gloas inherited-type audit

I reviewed the pre-existing containers that needed a Gloas counterpart rather than reuse of an older fork type.

**Directly changed by the Gloas spec:**
- `Attestation` and `IndexedAttestation` (from Electra): progressive container/list semantics.
- `BeaconBlockBody` (from Electra): progressive container/lists and the ePBS body changes.
- `BeaconState` (Fulu lineage): progressive container/lists and the Gloas state fields.
- `ExecutionPayload` (Deneb lineage): progressive lists/container plus Gloas fields.
- `ExecutionRequests` (from Electra): progressive lists and builder request fields.

**Replicated because they contain those changed types:**
- `AttesterSlashing` contains `IndexedAttestation`.
- `AggregateAndProof` and `SignedAggregateAndProof` contain `Attestation`.
- `BeaconBlock` and `SignedBeaconBlock` contain the Gloas body.

The direct changes are listed in the official Gloas spec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.12/specs/gloas/beacon-chain.md#modified-containers

`AttestationData` is deliberately different: its SSZ container and root are unchanged, but Gloas gives its existing `index` field a payload-availability meaning. It stays the shared Phase0 type and gets fork-specific validation instead of a duplicate Gloas type.

**Remaining confirmation limits:** the official `v1.7.0-alpha.12` static-vector corpus is not available to this checkout, so the whole set is not yet independently vector-verified. `AggregateAndProof` and `SignedAggregateAndProof` also were not present in the source fork; their shapes were checked against the spec and their roots have local tests, but they need the same vector confirmation. `ExpectedWithdrawals` is a spec transition dataclass with no matching client-side Go container here, so there is no type to replicate.

AntiD2ta:
## Integration reference — do not merge

This PR remains the unchanged integration reference for the original Gloas port. Please review and merge only the copied stack, in order:

1. #308 — Gloas type foundation
2. #309 — Gloas versioned API surface
3. #310 — ePBS block production and envelope transport
4. #311 — PTC and attestation compatibility
5. #312 — Gloas event forwarding
6. #313 — Gloas test harness and CI hardening

The copied stack includes one approved compatibility correction: `spec.BuilderVersion` and `spec.DataVersion` remain canonical types in `spec`, rather than moving to `spec/version`. The final PR also retains its scoped `.gitattributes` review metadata.

PR #309, opened 2026-08-05
feat: add Gloas versioned API surface

AntiD2ta:
## TL;DR

Adds Gloas arms to the versioned wrappers and the Gloas execution-payload envelope API surface.

## Dependency and review order

Review after the #308. Generated SSZ stays paired with its handwritten API type and generator manifest.

## Public API

Adds Gloas variants to versioned response wrappers, including signed proposal block-hash access.

## Behavior-sensitive checks

Fork dispatch and unknown-version errors are covered by wrapper tests.


api/versionedsignedproposal.go
```diff
@@ -496,6 +528,16 @@ func (v *VersionedSignedProposal) assertExecutionPayloadPresent() error {
 				return ErrDataMissing
 			}
 		}
+	case spec.DataVersionGloas:
```
Bez625, line 531: Do we not needed a concept of a `Blinded` block here?

api/v1/gloas/generate.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gloas
+
+//go:generate rm -f *_ssz.go
```
Bez625: It worries me how unbound this is. We've always explicitly deleted the ssz files by name rather than *. I think we should maintain that practice

spec/versionedsignedexecutionpayloadbid.go
```diff
@@ -0,0 +1,355 @@
...
+		DataVersionDeneb, DataVersionElectra, DataVersionFulu:
+		return ""
+	case DataVersionGloas:
+		if v.Gloas == nil {
+			return ""
+		}
+
+		return v.Gloas.String()
+	default:
+
```
Bez625: This blank line can go

Bez625 (changes requested):

api/versionedsignedproposal.go
AntiD2ta: Post-Gloas there is no blinded/unblinded block distinction: the proposal carries a signed execution-payload bid, and the payload is released later. The shared Blinded flag remains for earlier forks; the Gloas arm rejects it when set, and this is now covered by a regression test.

spec/versionedsignedexecutionpayloadbid.go
AntiD2ta: Fixed [80f743e](https://github.com/attestantio/go-eth2-client/pull/309/commits/80f743e8c6777edeb53785409cff82214113db63)

api/v1/gloas/generate.go
AntiD2ta: True, fixed [80f743e](https://github.com/attestantio/go-eth2-client/pull/309/commits/80f743e8c6777edeb53785409cff82214113db63)

Bez625 (approved):

PR #310, opened 2026-08-05
feat: add ePBS block production and envelope transport

AntiD2ta:
## TL;DR

Adds v4 ePBS block production plus execution-payload envelope retrieval and submission across HTTP, provider, mock, and multi-client layers.

## Dependency and review order

Review after the versioned API surface PR. Generated block-contents SSZ is paired with its handwritten type and manifest.

## Public API

Adds proposal and envelope options, provider interfaces, and `VersionedEPBSProposal`.

## Behavior-sensitive checks

Preset-aware SSZ request encoding, payload inclusion guards, envelope transport, and Gloas execution block-hash access.


AntiD2ta:
The ePBS submission transport tests now live in the dependent harness PR #313, where they use local HTTP handlers rather than accepting an arbitrary error. They assert handler receipt, POST method and route, `Content-Type`, `Eth-Consensus-Version`, `Eth-Blob-Data-Included` for envelopes, and a non-empty SSZ request body. The duplicate weak `ReachesServer` cases were removed from this PR so they cannot be mistaken for transport coverage.

Bez625 (approved):

PR #312, opened 2026-08-05
feat: add Gloas event forwarding

AntiD2ta:
## What this does

Adds seven Gloas SSE topics to the HTTP event client and forwards them through `multi.Events`. It also accepts those topics in `Event.UnmarshalJSON` without changing the existing `Event.Data` map behavior.

## Start with `multi/events.go`

The rewrite replaces thirteen handler methods with `newActiveHandler` and generic wrapper functions. It fixes a pre-existing forwarding bug: initially active clients received options without `Topics`, so their HTTP subscriptions failed. The retry path then subscribed with the caller's unwrapped options and forwarded events without active-address filtering.

Both subscription paths now use wrapped options. The wrappers preserve the caller's topics, forward only events from the active client, and leave handlers the caller did not supply nil so the generic handler can receive those topics. They also cover the previously missing block, block-gossip, and data-column-sidecar handlers.

## Gloas event types and decoding

`api.EventsOpts` gains seven topic-specific handlers. `api/v1` gains `ExecutionPayloadEvent`, `ExecutionPayloadGossipEvent`, `ExecutionPayloadAvailableEvent`, and `FastConfirmationEvent`. The bid, payload-attestation-message, and proposer-preferences topics use existing `spec/gloas` types.

The imported execution-payload event retains `ExecutionOptimistic`; the gossip event has no such field. Both require `builder_index` and `block_hash`. Their builder index uses `gloas.BuilderIndex`, and `gloas.BuilderIndexSelfBuild` names the self-build value.

Versioned Gloas payload handlers accept either a `{"version": ..., "data": ...}` wrapper or a bare object. `execution_payload_available` and `fast_confirmation` use plain JSON decoding.

## Tests

`multi/events_test.go` checks handler forwarding, generic fallback, active-client filtering, deferred activation, and ownership of caller-supplied topics. Node-independent HTTP tests cover every supported dispatch topic, Gloas handler precedence, wrapped and bare payloads, malformed input, and control cases. Tests also exercise both upstream execution-payload topic examples.

`api/v1/event_test.go` guards the topic allow-list against drift from `Event.UnmarshalJSON` and round-trips an `execution_payload_available` event through `Event` JSON.

The live blob tests skip a head without blob commitments. For a head with commitments, the blobs test requires a nonempty response; the sidecar test retains its existing skip for an empty response.

## Validation and limits

At `bab4e97`, Go 1.25.2 `gosilent test ./...` passes all 2,752 tests. The PR's GitHub test, lint, and revive checks pass.

Earlier direct-node Gloas HTTP tests passed against Prysm, Lighthouse, Teku, and Nimbus. After the gossip-type change, selected HTTP event tests were rerun against Prysm. These test results do not establish live SSE support from Teku: in the separate earlier observation, Teku returned an empty JSON response and closed the stream.

Outside the standard CI checks, the v1.7.0-beta.1 Gloas consensus vectors still fail on this branch's inherited YAML round-trip cases. The fix is in [PR #337](https://github.com/attestantio/go-eth2-client/pull/337), which is not part of this PR's head. A full local lint run also reported ten findings in unchanged files.


multi/events.go
```diff
@@ -52,29 +48,9 @@ func (s *Service) Events(ctx context.Context,
...
-		ah.opts.PayloadAttributesHandler = ah.payloadAttributesHandler
-		ah.opts.ProposerSlashingHandler = ah.proposerSlashingHandler
-		ah.opts.SingleAttestationHandler = ah.singleAttestationHandler
-		ah.opts.VoluntaryExitHandler = ah.voluntaryExitHandler
-
-		if err := client.(consensusclient.EventsProvider).Events(ctx, ah.opts); err != nil {
+		ah := newActiveHandler(s, log, client.Address(), opts)
+
+		if err := client.(consensusclient.EventsProvider).Events(ctx, ah.clientOpts); err != nil {
 			inactiveClients = append(inactiveClients, client)
```
AntiD2ta: **Events appends to a slice header copied from s.inactiveClients outside clientsMu, so it can write into the shared backing array that activateClient/deactivateClient also append into.**

activateClient builds s.inactiveClients with capacity len(active)+len(inactive), so it usually has spare capacity. When an active client's Events() fails, `inactiveClients = append(inactiveClients, client)` writes slot len(s.inactiveClients) of the shared array with no lock held. A concurrent deactivateClient under the lock does `append(s.inactiveClients, x)` into the same slot. One write overwrites the other: this Events call retries the wrong client (the one that failed is never subscribed), or s.inactiveClients silently takes the wrong entry. The race detector would flag it. Pre-existing, but it is in the function this PR rewrites. Fix: take `slices.Clone(s.inactiveClients)` under the RLock.

multi/events.go
```diff
@@ -107,19 +79,21 @@ func (s *Service) Events(ctx context.Context,
...
 					return
 				}
 
 				if !syncResponse.Data.IsSyncing {
-					// Client is now synced, set up the events call.
-					if err := c.(consensusclient.EventsProvider).Events(ctx, opts); err != nil {
+					// Client is now synced, set up the events call.  This uses the same substituted
+					// options as an initially-active client, so that events from it are subject to
+					// the same active-address filtering.
+					if err := c.(consensusclient.EventsProvider).Events(ctx, ah.clientOpts); err != nil {
```
AntiD2ta: **A deferred client whose NodeSyncing check or Events() call fails once is never subscribed, but it can still become the primary later, and then the new active-address filter drops every event.**

Clients A (primary) and B (not synced at subscribe time). B's goroutine gets one transient NodeSyncing error (or its Events() fails) and returns for good. Later A falls out of sync and multi promotes B to primary. Now s.Address()==B, so A's stream is filtered out. B has no stream at all. The caller gets no more events, and nothing tells them. Before this PR the late path was unfiltered, so this silent stall was not possible there. Retry instead of returning, or subscribe every client regardless of its sync state.

http/events_internal_test.go
```diff
@@ -14,187 +14,399 @@
...
-func TestEventHandler(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	handled := false
-	handler := func(*apiv1.Event) {
-		handled = true
-	}
-
+func TestHandleEventDispatchesEverySupportedTopic(t *testing.T) {
```
AntiD2ta: **The dispatch test swapped real node payloads (head, block, attestation, attester_slashing, chain_reorg, proposer_slashing, bls_to_execution_change, contribution_and_proof) for data built by marshalling the library's own zero-valued types, so it can no longer catch a wire-format mismatch.**

Say a type's MarshalJSON and UnmarshalJSON drift together away from what beacon nodes emit, for example a renamed key or a changed bitfield encoding. The new test marshals with the library and decodes with the library, so it still passes, while every real SSE message fails to parse and gets logged and dropped. The deleted cases used verbatim mainnet payloads and would have failed. Keep at least one real-wire fixture per topic, like the spec-example test added for execution_payload.

multi/events_test.go
```diff
@@ -54,3 +67,347 @@ func TestEvents(t *testing.T) {
...
+}
+
+// TestEventsOwnsTopicsForDeferredClients confirms that a later sync check retains the requested
+// topics when the caller reuses the options slice after Events returns.
+func TestEventsOwnsTopicsForDeferredClients(t *testing.T) {
+	ctx := context.Background()
+
+	logs := make(chan string, 1)
+	originalLogger := zerologger.Logger
+	zerologger.Logger = zerolog.New(writerFunc(func(data []byte) (int, error) {
```
AntiD2ta: **The test replaces the global zerolog logger with a writer that blocks on a 1-slot channel. multi.New captures that logger, and its monitor goroutine (running on context.Background) keeps it after the test restores the global.**

After the test returns nobody drains `logs`. Any second error-level line from this multi instance (the monitor or recheck path, or a retry goroutine) blocks forever inside Write, which leaks the goroutine for the rest of the test binary. The global swap also races with any still-running goroutine from an earlier test that reads zerologger.Logger. Use a non-blocking send or a buffered sink, and cancel a ctx passed to multi.New in t.Cleanup.

http/events_internal_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2020, 2021 Attestant Limited.
+// Copyright © 2026 Attestant Limited.
```
AntiD2ta: **The copyright header of a file that dates from 2020 was changed from `2020, 2021` to just `2026`, which drops the original years.**

The git history shows this file was created on 2020-12-17. Replacing the header with `Copyright © 2026` misstates when the work was first copyrighted. Every other touched file in this PR extends its range (for example `2020 - 2026`). This one should read `2020 - 2026` too.

api/v1/event.go
```diff
@@ -33,23 +34,35 @@ type Event struct {
 	Data any
 }
 
-// SupportedEventTopics is a map of supported event topics.
+// SupportedEventTopics is a map of supported event topics. It is the allow-list
```
AntiD2ta: **Each topic is still listed in five places that must stay in sync (SupportedEventTopics, Event.UnmarshalJSON, checkEventSpecificHandler, handleEvent, and newActiveHandler), and the PR adds tests to catch drift instead of removing the duplication.**

Adding one Gloas topic took edits to five switches and maps across three packages, plus two reflection or cross-check tests to catch a missed one. The next fork's topics need the same five edits. A single per-topic registry (topic -> new(T), handler accessor) would produce the allow-list, the unmarshal switch, the handler check and the dispatch from one table.

http/events.go
```diff
@@ -623,3 +652,208 @@ func (*Service) handleDataColumnSidecarEvent(ctx context.Context,
 		log.Debug().Msg("No specific or generic handler supplied; ignoring")
 	}
 }
+
+func (*Service) handleExecutionPayloadEvent(ctx context.Context,
```
AntiD2ta: **The seven new handleXxxEvent functions are 27-line copies that differ only in the data type, the decoder, the specific-handler field and the log text.**

About 190 duplicated lines are added on top of the 15 existing copies. A fix to the dispatch or fallback logic now has to be applied 22 times. multi/events.go in this same PR already uses a generic `substitute[T]`. A `dispatchEvent[T any](ctx, msg, decode, specific func(context.Context, T), generic)` helper would cut each topic to one line.

http/events.go
```diff
@@ -623,3 +652,208 @@ func (*Service) handleDataColumnSidecarEvent(ctx context.Context,
...
+		})
+	default:
+		log.Debug().Msg("No specific or generic handler supplied; ignoring")
+	}
+}
+
+// unmarshalVersionedEventData decodes an SSE event payload that the beacon-API
+// spec wraps as {"version": "...", "data": {...}}. Some clients (Prysm) still
+// send the bare unwrapped object, so that shape is accepted as a fallback.
+func unmarshalVersionedEventData(raw []byte, v any) error {
```
AntiD2ta: **unmarshalVersionedEventData's comment says the spec wraps these payloads in {version,data}, but the spec example for execution_payload and execution_payload_gossip is bare. Every spec-compliant message for those topics is also decoded twice.**

The beacon-APIs eventstream examples for execution_payload and execution_payload_gossip are bare objects. Only bid, payload_attestation_message and proposer_preferences are wrapped. So for the two payload topics the fallback path is the spec path. The comment calls it a Prysm workaround, which will mislead the next person who edits it. Each spec-compliant payload event is also scanned once into the wrapper and then parsed again. Use plain json.Unmarshal for the two bare topics, or fix the comment.

api/v1/event_test.go
```diff
@@ -111,3 +115,57 @@ func TestEvent(t *testing.T) {
...
 }
+
+// TestSupportedEventTopicsGloas confirms the allow-list carries the event
+// topics introduced by the Gloas (ePBS) fork. SupportedEventTopics is what the
+// http client gates every Events() subscription on, so a topic missing from it
+// is unreachable from the public API however completely the rest of the client
+// handles it.
+//
+// This is the guard that runs unconditionally: the http package's tests are
+// skipped wholesale unless HTTP_ADDRESS names a beacon node, so none of them
```
AntiD2ta: **The new comment says http tests are 'skipped wholesale unless HTTP_ADDRESS names a beacon node', but http/main_test.go in this same PR now runs them without an address.**

This PR changes TestMain to call m.Run() without HTTP_ADDRESS, and testService now calls t.Skip per test. So node-free http tests, including the new events_internal_test.go dispatch tests, do run in CI. The justification in the comment is false the moment this PR merges and will mislead future maintainers about where the coverage lives.

api/v1/fastconfirmationevent.go
```diff
@@ -0,0 +1,98 @@
...
+	Block       string `json:"block"`
+	CurrentSlot string `json:"current_slot"`
+}
+
+// MarshalJSON implements json.Marshaler.
+func (e *FastConfirmationEvent) MarshalJSON() ([]byte, error) {
+	return json.Marshal(&fastConfirmationEventJSON{
+		Slot:        fmt.Sprintf("%d", e.Slot),
+		Block:       fmt.Sprintf("%#x", e.Block),
+		CurrentSlot: fmt.Sprintf("%d", e.CurrentSlot),
```
AntiD2ta: **UnmarshalJSON treats a missing current_slot as optional, but MarshalJSON always writes it as "0", so an event from a client that omits the field re-marshals with a made-up current_slot of 0.**

A client that does not emit current_slot sends {"block":...,"slot":"100"}. The event decodes with CurrentSlot=0. The generic handler path and Event.MarshalJSON then emit "current_slot":"0", which says the algorithm ran at genesis and is older than the confirmed slot 100. Downstream consumers cannot tell a missing value from slot 0. Model it as *phase0.Slot, or omit the key on marshal when it was absent.

multi/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/d9ef82a7e2b6b6b9abb31f1a6c513418513e64a0. `Events` now takes `slices.Clone(s.inactiveClients)` under the read lock. `TestEventsDoesNotWriteServiceClientLists` gives the service's inactive list spare capacity and fails if `Events` writes into it.

multi/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/eeeb6892d696db91104be7588a79d6cc8bb2ca91. A failure to fetch sync state or to subscribe is now logged as a warning and retried at the interval, until the subscription's context is done. `TestEventsRetriesDeferredClient` covers both failures, and `TestEventsStopsRetryingOnContextDone` checks that the retries stop.

http/events_internal_test.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/de39efd3edd816d9a7be0c2d9d9dece44ac5361d. The dispatch table decodes real payloads again: the ten mainnet payloads the old test had, plus beacon-APIs eventstream examples for the other topics. This showed that the library can't yet decode master's Gloas examples for `data_column_sidecar` and `payload_attributes`, or v4.0.0's Electra `payload_attributes`. Those topics use the latest example it can decode, and the test comment records why. The spec's 96-byte KZG commitments are cut to 48.

multi/events_test.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/eeeb6892d696db91104be7588a79d6cc8bb2ca91. The test no longer touches the global logger. It checks the topics the deferred client is subscribed with.

http/events_internal_test.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/de39efd3edd816d9a7be0c2d9d9dece44ac5361d. The header is now `2020 - 2026`.

api/v1/event.go
AntiD2ta: Partly addressed in https://github.com/attestantio/go-eth2-client/pull/312/commits/dc322e87b31cd64d54fd1bc1bc3b49cdaa257089 and https://github.com/attestantio/go-eth2-client/pull/312/commits/785d745eebe8f6bf09216600a875fb2ca086ba66, which take the five lists down to two. `api/v1` now keeps one topic-to-payload map, and both `SupportedEventTopics` and `Event.UnmarshalJSON` come from it, so the drift test is gone. In `http`, one table drives both the handler check and dispatch. The `http` table remains separate because its decoders and payload types differ from `api/v1`'s (for example, `attester_slashing` decodes as electra there and as phase0 in `api/v1`). `multi`'s substitution list is per handler field, not per topic, and `TestEventsForwardsEveryHandler` walks the fields reflectively to guard it.

http/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/dc322e87b31cd64d54fd1bc1bc3b49cdaa257089. The 22 `handleXxxEvent` functions are replaced by a generic `newEventTopic` and one table entry per topic.

http/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/b63341df13c15b4aa14572d78f2c9d28b1c335d0. `execution_payload` and `execution_payload_gossip` now decode with `json.Unmarshal`. The comment now names the three topics the spec wraps: `execution_payload_bid`, `payload_attestation_message` and `proposer_preferences`.

api/v1/event_test.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/785d745eebe8f6bf09216600a875fb2ca086ba66. The comment now says the http dispatch test also pins the list, and that this test is kept as a guard next to the list that doesn't depend on the http client.

api/v1/fastconfirmationevent.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/83ec9afe62f05bb76985a218d6d7304c3e78982e. `CurrentSlot` is now a `*phase0.Slot`: nil when the node omits the field, and omitted again on marshal. The no-`current_slot` case is a round-trip row in the table test.

multi/events.go
```diff
@@ -83,52 +62,58 @@ func (s *Service) Events(ctx context.Context,
...
+					// the same active-address filtering.
+					err := c.(consensusclient.EventsProvider).Events(ctx, ah.clientOpts)
+					if err == nil {
+						ah.log.Trace().Str("address", ah.address).Strs("topics", ah.clientOpts.Topics).Msg("Events handler active")
+
+						return
+					}
 
-					return
+					ah.log.Warn().
```
AntiD2ta: **Permanent subscribe errors now retry forever, and `multi.Events` still returns nil.** The old retry path gave up after one attempt ("Return either way"). This loop now retries every error from `Events`, including validation errors that can never succeed: `no topics supplied` (`ErrInvalidOptions`), `unsupported event topic X`, and `no handler for X event`. Example: a caller passes `Topics: nil` or a misspelled topic to `multi.Events`. Each HTTP client rejects it, the initially-active path moves the client to `inactiveClients`, `multi.Events` returns `nil`, and one goroutine per client logs `Failed to set up events handler; will retry` every 5s until ctx ends. For a long-lived ctx, that is never. Either validate `opts` once up front and return the error to the caller, or stop retrying when `errors.Is(err, client.ErrInvalidOptions)` (the http topic and handler errors would need that sentinel too).

multi/events_internal_test.go
```diff
@@ -0,0 +1,62 @@
...
+		return errors.New("failed to subscribe")
+	}
+
+	inactive, err := mock.New(ctx, mock.WithName("inactive"))
+	require.NoError(t, err)
+
+	inactiveClients := make([]consensusclient.Service, 1, 2)
+	inactiveClients[0] = inactive
+
+	s := &Service{
```
AntiD2ta: **This `Service` literal leaves `eventsRetryInterval` at zero, so the retry goroutine busy-spins for the rest of the test binary.** `failing` is not syncing (mock `SyncDistance` 0), so the goroutine calls `Events`, gets `failed to subscribe`, and then waits on `time.After(0)`. It does this with `context.Background()`, which is never cancelled. The result is a tight loop that burns a core under `-race` for every later test in `./multi/`. Set `eventsRetryInterval` (and use a cancellable ctx with `t.Cleanup(cancel)`). Also consider defaulting a zero interval inside `Events` so a zero-value `Service` cannot spin.

http/events.go
```diff
@@ -181,445 +144,128 @@ func (s *Service) handleEvent(ctx context.Context,
...
-	case opts.VoluntaryExitHandler != nil:
-		opts.VoluntaryExitHandler(ctx, data)
-	case opts.Handler != nil:
-		opts.Handler(&apiv1.Event{
-			Topic: string(msg.Event),
-			Data:  data,
-		})
-	default:
-		log.Debug().Msg("No specific or generic handler supplied; ignoring")
+	if err := json.Unmarshal(raw, &wrapper); err == nil && len(wrapper.Data) > 0 && wrapper.Version != "" {
```
AntiD2ta: **`unmarshalVersionedEventData` discards `version` without checking it.** Every `execution_payload_bid`, `payload_attestation_message` and `proposer_preferences` event is decoded into the fixed Gloas type whatever version it declares. After the next fork changes one of these containers, a node sending `{"version":"<next fork>","data":{...}}` gets decoded as Gloas. Unknown fields are silently ignored, and the handler gets a wrong but valid-looking object with no error logged. The version is also not surfaced to generic-handler consumers. At minimum, reject a non-`gloas` version (so the event fails loudly rather than being mis-decoded), or return a versioned wrapper as `payload_attributes` does.

api/v1/event.go
```diff
@@ -33,23 +34,46 @@ type Event struct {
...
-	"proposer_slashing":       true,
-	"single_attestation":      true,
-	"voluntary_exit":          true,
+// eventTopicData is the set of supported event topics, each with a function returning a new
+// value of the type into which the data of an event of that topic decodes.  It is the single
+// list of topics in this package: SupportedEventTopics and Event.UnmarshalJSON are both derived
+// from it.
+var eventTopicData = map[string]func() any{
+	"attestation":                 func() any { return &spec.VersionedAttestation{} },
+	"attester_slashing":           func() any { return &phase0.AttesterSlashing{} },
```
AntiD2ta: **`eventTopicData` claims to be "the single list of topics", but `http/events.go` keeps a second per-topic table with different decode types.** `attester_slashing` decodes to `*phase0.AttesterSlashing` here but to `*electra.AttesterSlashing` in `http.eventTopics`. So a generic handler sees `*electra.AttesterSlashing` from the live stream and `*phase0.AttesterSlashing` after an `Event` JSON round-trip. The two tables will keep drifting: every new topic needs matching edits here, in `http.eventTopics`, and in `multi.newActiveHandler`. Tests pin the key sets, not the types. Derive one from the other, or at least assert that the types match.

http/events.go
```diff
@@ -124,52 +119,20 @@ func (s *Service) checkEventsOpts(opts *api.EventsOpts) error {
...
-	case "payload_attributes":
-		hasHandler = opts.PayloadAttributesHandler != nil
-	case "proposer_slashing":
-		hasHandler = opts.ProposerSlashingHandler != nil
-	case "single_attestation":
-		hasHandler = opts.SingleAttestationHandler != nil
-	case "voluntary_exit":
-		hasHandler = opts.VoluntaryExitHandler != nil
-	default:
+	handling, exists := eventTopics[topic]
```
AntiD2ta: **`checkEventsOpts` validates each topic twice, against two maps that are meant to be identical.** It checks `apiv1.SupportedEventTopics` first, then `eventTopics` here. The `unsupported event %s` branch is reachable only when the two maps disagree. When they do disagree and a generic handler is set, this branch is skipped entirely: the subscription is accepted, and every event of that topic then logs `Received message with unhandled topic` at Warn. Validate against `eventTopics` alone, which is what `handleEvent` dispatches on. `SupportedEventTopics` then stays only as a guard in tests.

multi/events.go
```diff
@@ -140,215 +125,110 @@ type activeHandler struct {
...
+	sub := &api.EventsOpts{
+		Common: opts.Common,
+		Topics: slices.Clone(opts.Topics),
+	}
+	ah.clientOpts = sub
+
+	// These need no nil check of their own: substitute leaves a handler the caller did not
+	// supply nil, for the reason given on it.
+	sub.Handler = substituteGeneric(ah, opts.Handler)
+	sub.AttestationHandler = substitute(ah, "attestation", opts.AttestationHandler)
```
AntiD2ta: **This is a third hand-maintained topic→handler list, and its topic strings are unchecked.** The comment calls `topic` "diagnostic only", so a copy-paste slip here, such as passing "head" for `BlockHandler`, mislabels every trace line and nothing catches it. The same field-by-field mapping already exists as closures in `http.eventTopics`. A shared per-topic descriptor in `api` would let `http`, `multi` and `api/v1` iterate one table instead of three: topic name, a getter, and a setter or wrapper for the handler field. The reflective `TestEventsForwardsEveryHandler` only proves that each field is wired, not that it has the right topic.

multi/events.go
```diff
@@ -83,52 +62,58 @@ func (s *Service) Events(ctx context.Context,
...
-						Strs("topics", opts.Topics).
+						Strs("topics", ah.clientOpts.Topics).
 						Err(err).
-						Msg("Failed to obtain sync state from node")
+						Msg("Failed to obtain sync state from node; will retry")
+				case !syncResponse.Data.IsSyncing:
+					// Client is now synced, set up the events call.  This uses the same substituted
+					// options as an initially-active client, so that events from it are subject to
+					// the same active-address filtering.
+					err := c.(consensusclient.EventsProvider).Events(ctx, ah.clientOpts)
```
AntiD2ta: **Unchecked type assertion `c.(consensusclient.EventsProvider)` inside a goroutine.** If a configured client (e.g. a `testclients` wrapper or a custom `Service`) implements `NodeSyncingProvider` but not `EventsProvider`, this panics in a background goroutine and crashes the process. The sync-provider check just above uses the two-value form and logs. Use the comma-ok form here and at line 56 too. That line panics synchronously for the same client.

api/v1/fastconfirmationevent.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	if err := decodeFixedBytes(e.Block[:], data.Block, "block"); err != nil {
+		return err
+	}
+
+	// current_slot was added to the spec after slot/block; parse it when
+	// present but tolerate clients that do not yet emit it, leaving it nil
+	// rather than 0 so that it is not re-marshalled with a made-up value.
+	if data.CurrentSlot != "" {
```
AntiD2ta: **`UnmarshalJSON` does not clear `CurrentSlot` when `current_slot` is absent.** When a `FastConfirmationEvent` value is reused, e.g. decoding successive events into the same struct, a payload without `current_slot` keeps the previous event's `CurrentSlot`. That reports a stale wall-clock slot and contradicts the doc comment ("or nil if the node did not supply it"). Set `e.CurrentSlot = nil` before the `if`.

multi/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/6e4a94dab6ce62c048709141b922bc4336e292bd. `multi.Events` now validates the options with `http.ValidateEventsOpts` before calling any client, so nil options, empty or unsupported topics and missing handlers are returned to the caller wrapped in `client.ErrInvalidOptions`, and never reach the retry loop. `TestEventsRejectsMissingHandler`, `TestEventsRejectsUnknownTopic` and `TestEventsRejectsEmptyTopics` cover each case.

multi/events_internal_test.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/7a505a0c30c282535ff831d8c46158c645834779. The retry loop falls back to the 5s default when `eventsRetryInterval` is not positive, so a `Service` not built by `New` no longer spins. This test now sets an interval and ends its context with the test. `TestEventsRetriesWithDefaultIntervalWhenUnset` checks that an unset interval waits between attempts.

http/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/c7b39da143727766952b511b2928c5a1b4827def. Each versioned topic now records its fork, and `EventTopic.Decode` rejects a wrapped payload whose `version` is not that fork, instead of decoding it into the Gloas type. Bare payloads are still accepted. `TestEventTopicDecode` covers the wrapped, bare, other-fork and unknown-fork cases.

api/v1/event.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/0759ce6ad3146fba4ac7b90a64cb0047f080131a. Each topic is now described once in `api/v1` as an `EventTopic[T]`, and `SupportedEventTopics`, `Event.UnmarshalJSON`, http's dispatch and multi's filtering are all derived from it. `attester_slashing` decodes to `*electra.AttesterSlashing` everywhere. The binding to `EventsOpts` fields is generic, so a handler that does not take the topic's type does not compile. `TestEventTopicDecodeTypes` pins each topic's name and decode type, not just the key set.

http/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/0759ce6ad3146fba4ac7b90a64cb0047f080131a. http no longer has a table of its own: `ValidateEventsOpts` checks each topic against `SupportedEventTopics` once, then asks `opts.HasTopicHandler` for a specific handler when there is no generic one. The dead 'unsupported event' branch is gone. `TestEventsOptsBindsEveryTopicHandler` checks that every supported topic is bound, so the two cannot drift.

multi/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/0759ce6ad3146fba4ac7b90a64cb0047f080131a. The per-field `substitute` calls and their topic strings are gone. `newActiveHandler` now calls `opts.Filtered(ah.forwards)`, which wraps every handler through the single binding list in `api`, using the topic name from the descriptor. `TestEventsForwardsEveryHandler` still walks every `EventsOpts` handler field reflectively.

multi/events.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/81b820e6b6d77b0724df54f44afc05be5a1790cb. Both assertions, for active clients and in the retry goroutine, now use the comma-ok form, and a client that does not provide events is logged and skipped. `TestEventsSkipsClientsWithoutEvents` passes such a client as both active and deferred, and fails with a panic without the fix.

api/v1/fastconfirmationevent.go
AntiD2ta: Fixed in https://github.com/attestantio/go-eth2-client/pull/312/commits/6440a0c5997a291d6939a82b39092116219f35e7. `UnmarshalJSON` resets `CurrentSlot` to nil before parsing `current_slot`. `TestFastConfirmationEventJSON/CurrentSlotAbsentOnReuse` decodes an event with `current_slot` and then one without into the same value, and fails without the fix.

PR #323, opened 2026-09-02
feat: use BuilderConfig POST for Gloas block production

AntiD2ta:
## Summary

Migrate Gloas block production to the POST `produceBlockV4` contract.

The client now sends a required `BuilderConfig` to
`/eth/v4/validator/blocks/{slot}` and negotiates JSON or SSZ. It retains
builder routing metadata, supports both valid response containers, and rejects
responses whose bid, payload, or block commitments disagree.

## Review guide

Start with `http/epbsproposal.go`. It builds the POST request, selects the
response codec, and checks the block, envelope, builder index, and
execution-request commitments.

`spec/gloas/builderconfig.go` and its generated SSZ files define the request
types. `builderconfig_json.go` handles JSON encoding and bounds checks.

`api/epbsproposalopts.go`, `api/versionedepbsproposal.go`, and
`api/submitproposalopts.go` contain the public API changes.

## Behaviour and compatibility

- Replace `BuilderBoostFactor` with a required `BuilderConfig`. An empty
  builders list requests no direct builder bids.
- Send POST only. The client does not retry the legacy GET route.
- Validate each builder authorization against the requested slot before sending
  it, so stale authorizations fail locally instead of silently losing a bid.
- A builder-backed response may contain a bare block when `include_payload=true`.
  Callers must inspect `ExecutionPayloadIncluded`.
- Missing execution value remains unknown. `VersionedEPBSProposal.Value()`
  returns `nil` when the execution-value header is absent.
- Preserve `Eth-Builder-Url` in response metadata and forward an optional
  `BuilderURL` when publishing the signed block.
- Require `Eth-Consensus-Version` on every response. SSZ responses also require
  `Eth-Execution-Payload-Included`.

## Validation

- Node-free JSON and SSZ request and response tests cover headers, bodies,
  metadata, bounds, response limits, commitment mismatches, and custom presets.
- `go generate ./spec/gloas`
- `gosilent test ./api ./spec/gloas`
- Focused HTTP protocol tests
- `go build ./...`

### Live node runs

Run on 2026-09-09 against the direct Glamsterdam devnet-8 beacon nodes, one
client at a time:

```
HTTP_ADDRESS=... HTTP_REQUIRE_GLOAS=true HTTP_TEST_CONCURRENCY=1 \
  gosilent test -timeout=15m -p 1 -run 'Gloas|EPBS|ExecutionPayload' ./http/...
```

**Prysm `v7.1.8-6aea591`** (`develop`, POST-only): the whole selection passed,
139 passed and 0 failed. Both `BuilderConfig` shapes reached the node in both
encodings: an empty local-preferred list
(`SSZ`/`JSONPayloadExcluded`, `SSZ`/`JSONPayloadIncluded`) and a populated
single-entry list (`SSZ`/`JSONBuilderRequested`). `TestSubmitProposalGloas`
then published the produced block through `/eth/v2/beacon/blocks`.

The same selection against Prysm `v7.1.8-81a701c` failed exactly those eight
tests with `405 Method Not Allowed`. That commit predates POST support, which
arrived in prysmaticlabs/prysm#17392 and became POST-only in
prysmaticlabs/prysm#17397. The method probe inverts across the upgrade:

| | `81a701c` | `6aea591` |
|---|---:|---:|
| `GET /eth/v4/validator/blocks/{slot}` | 400 | 405 |
| `POST /eth/v4/validator/blocks/{slot}` | 405 | 400 |

**Lighthouse:** the six `TestEPBSProposal` rows and `TestSubmitProposalGloas`
passed. Observed 2026-09-08, before the devnet client images were refreshed.

**Teku and Nimbus:** neither was green for reasons outside this change. Teku
answered the POST route but rejected the test's `headSlot + 1` target as a
historic slot. Nimbus has no v4 block-production route and returned 404.
Neither client has been re-probed since the 2026-09-09 image refresh.


AntiD2ta:
This PR was merged accidentally while its changes were not present on #316. Its replacement is #334, which preserves this PR’s original head commit as the corresponding layer in GitHub stack #336.

